Issue #29, opened 2018-08-07
Failing examples for removing comments


kylebarron:
![image](https://user-images.githubusercontent.com/15164633/43784511-4d940da0-9a32-11e8-8a54-4de13db8894d.png)

kylebarron:
https://gist.github.com/kylebarron/e46ce79911976c3aae9d7e98309b9222

kylebarron:
@mcaceresb Check out https://github.com/kylebarron/stata_kernel/blob/dev-stata-lexer/stata_kernel/stata_lexer.py

kylebarron:
I believe I've implemented correct comments. Working on strings then writing some tests.

kylebarron:
Yes that should be possible. Since Mata is only started with `mata` and ended with `end` (is this true?), I can easily find when Mata starts and ends and then just write a different lexer for Mata and use the Mata tokenizer on the Mata code.

mcaceresb:
You need to be careful in that

1. `mata` can be preceded by any number of `CAPture`, `QUIetly`, or `Noisily`
2. `mata` can be followed by any number of spaces, `:`, and any number of spaces thereafter.
3. If there is anything else following mata, it is evaluated inline, and does not enter the mata environment
4. `end` can be preceded by any number of spaces and followed by __anything__
5. It doesn't matter where `end` appears. In a loop, a function, etc. The mata environment will end at that point.
6. `mata` is not nested, as `mata` itself is a valid `mata` command.

kylebarron:
Thanks. I'll have to look at the mata syntax more closely.

mcaceresb:
Comments are just C-style comments, except they can be nested. Rules for locals, etc. are the same. I don't think there is much to look out for within a mata block per se, other than the prompts are different.

kylebarron:
I'll come back to Mata once I progress on some other stuff. You're welcome to submit a PR with a Mata lexer if you'd like. All it needs to support is comments and strings.

PR #36, opened 2018-08-08
Use Pygments lexer to remove comments

kylebarron:
Fixes https://github.com/kylebarron/stata_kernel/issues/29

@mcaceresb

image

### DIFF
--- a/stata_kernel/code_manager.py
+++ b/stata_kernel/code_manager.py
@@ -1,0 +1,19 @@
+from pygments import lex
+
+from .stata_lexer import StataLexer
+
+
+class CodeManager(object):
+    def __init__(self, code):
+        self.input = code
+        self.tokens = self.tokenize_code(code)
+
+    def tokenize_code(self, code):
+        lexer = StataLexer(stripall=True)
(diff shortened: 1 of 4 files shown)


mcaceresb:
The lexer is broken for me, atm. I think you are not joining back code chunks correctly? It seems each chunk is delimited by a space or a line break, but you are treating them as lines. At least for me, something like `disp "hi"` evaluates `disp` and then `"hi"`.

See [these lines](https://github.com/kylebarron/stata_kernel/blob/master/stata_kernel/kernel.py#L270-L280). Your comment is not correct because `strip` also removes line breaks. Not sure that `strip(' ')` is required either, tho.

However, if you look at a `chunk` you will see a "word", not a line.

mcaceresb:
This is what happens for me

```python
disp "hi"
"hi is not a valid command name
r(199);
```

kylebarron:
On linux?

mcaceresb:
Yes.

kylebarron:
Can you try adding a space at the end

mcaceresb:
Same.

PR #44, opened 2018-08-09
[WIP] Stable API

### LINKED ISSUE #3: Windows support
First of all, awesome work! I will try this kernel out on my Linux workstation soon and let you know if I run into any issues. 

I have been doing a little bit of research into Windows support but it seems that the Windows version does not have an interactive console mode (only the full GUI version or the console batch mode). It does appear that it is possible to install Stata without GUI, but I have not been able to get this to work yet (e.g. https://www.stata.com/support/faqs/windows/install-from-command-line/). 

Another option would be to include Stata Automation integration for Windows users (like I did with ipystata). I can probably add this for you once I get back from holiday in a couple of weeks. 

I will also definitely consider adding similar Linux (and MacOS) functionality into ipystata, by the looks of it this approach using `pexpect` seems to work nicely.

### LINKED ISSUE #21: Stata 14 support
I suspect that the `nomsg` parameter was added with the most recent version Stata (i.e. 15). 

When trying to execute code using Stata 14 I get an error:

image

https://github.com/kylebarron/stata_kernel/blob/2d408bcd70188bc8e0d592eff10db948f52f0ff9/stata_kernel/kernel.py#L297

### LINKED ISSUE #28: Fix blocks: cap, qui, noi, if, else, else if
image

### LINKED ISSUE #32: code_manager.py file
This file should have a class that accepts an unmodified input string and returns a list of run-able code blocks. In general each item of the list should be a single line of Stata code, except when that line of Stata code would not return to a dot prompt.

```py
class CodeManager(object):
    def __init__(self, code):
        self.input = code
        self.tokens = self.tokenize_code()

    def tokenize_code(self):
        # Do tokenizing with Pygments
        # Return a list of tokens

    def is_complete(self):
        # Analyze tokens
        # Determine if all of input is complete.
        # Return simple True/False

    def completions(self, cursor_pos):
        # Unclear whether it's necessary to tokenize the code for this.
        # Return dict with correct Jupyter message

    def tokens_to_lines():
        # Form logical code chunks from tokens
        # This should also remove comments
        # Returns a list of code chunks that are safe to run in the console
```

### LINKED ISSUE #38: Tokens issue
@mcaceresb

The specific issue here is that strings are tokenized differently than text. With my current code joining method, I was thinking there would be two different Token types, `Token.Text` and `Token.MatchingBracket.Other`, where the latter is a contiguous set of tokens that comprise a block (and which thus must be sent all at once). 

Obviously I forgot that I had also defined the token type `Token.Literal.String`. Since this is a different name than the earlier two token types, the loop parsed it into a different line.

I need to leave the string token type so that it catches comments, and also within blocks:
image

This is the part where it's superbly annoying that there isn't a way with Pygments to show the entire token stack. It only shows the topmost token, so I can't see (I don't think) that above the `Token.Literal.String.Double` is inside a `Token.MatchingBracket.Other`. 

I think the main way to solve this is to set the token name for strings that occur within a block to `Token.MatchingBracket.Other`

### LINKED ISSUE #39: When code sent to do_execute is not complete, return invalid response
The `do_is_complete` function prevents the Jupyter kernel from sending input until there's a valid, complete chunk of code. But when Atom sends code, there's no way to probe for more user input, so the selection goes directly to `do_execute`. Because of this, I can't assume that the code sent to `do_execute` is necessarily complete, and I need to check.

### LINKED ISSUE #41: Text wrapping with long input/output using shell
image

### LINKED ISSUE #43: Use single cache directory for logs, images
#### Code Sample, a copy-pastable example if possible

```python
# Your code here

```
#### Problem description

[this should explain **why** the current behaviour is a problem and why the expected output is a better solution.]

**Note**: Many problems can be resolved by simply upgrading `medicare_utils` to the latest version. Before submitting, please check if that solution works for you.

### LINKED ISSUE #52: Add line continuation indent to loop lines
1. image
   image

---

2. Have to add the line continuation spaces after the `. `.
image
image

kylebarron:
Basically a rewrite of most things. There are a lot of thoughts in the docstrings.

Important things to keep in mind:
- Text wrapping. #41
- Though I don't have to add it now, make sure the design is compatible with Mata and `#delimit ;`

Fixes #32, fixes #28, fixes #38, fixes #39, fixes #41, fixes #43, fixes #3,  fixes #52, fixes #21.


kylebarron:
@mcaceresb Can you test dev now?

I think the current structure in `dev` is stable...

kylebarron:
Yay. Png/pdf images work now too. It was hard to figure out cause there's no documentation, but [this issue](https://github.com/jupyter/jupyter_client/issues/2) gave the clue that you needed to encode it as base64 (though it wasn't clear that then you still needed to decode it as utf-8 to transmit the image).

Issue #24, opened 2018-08-07
Delete .stata_kernel_images folder on kernel shutdown


kylebarron:

kylebarron:
Yes I agree, but I was thinking of the dua model, and I'm not sure it would be ok to arbitrarily save images in tmp (likewise with the cross-console history file, not sure that can go in ~ by default. Maybe I could put this setting in the configuration.

mcaceresb:
The "tmp" folder created via `tempfile` is only readable and writeable by the user that makes it, AFAIK. I did ask about this when I started work because the DUA model says all the things must be in the user's DUA folder. However, SAS and Stata create temporary files in the system's temporary directory all the time. I was told that was OK.

Nevertheless, an option to specify the images cache directory would certainly be nice.

kylebarron:
I'm not sure it's necessary to delete the cache directory anymore, as long as it's always in the same place and now possible to be user defined.

PR #49, opened 2018-08-10
Added hack-ish version of magics: %plot, %locals, %globals

mcaceresb:
- [ ] tests added / passed (not sure what the format is for these)
- [ ] whatsnew entry (did not find "whatsnew")

%plot [--width width] [--scale height] [--scale SCALE] code

This displays the plots made by Stata with height and width as
specified, scaled by SCALE. This makes it feasible for the user to
request a plot be displayed even if it is embedded in a program or a
fenced block. `code` and also print anything and the plot will still be
displayed (so this overrides the Hydrogen compatibility check).

%globals [-t | --truncate] [regexp?]

Print all globals matching `regexp?` and their contents. If `regexp` is
not specified, all globals are printed. If `--truncate` is specified,
only the first line of each global's contents, as truncated by Stata's
`macro dir`.

%locals [-t | --truncate] [regexp?]

As above, but for locals.

This also moves graphs by default to a temp folder, as noted in https://github.com/kylebarron/stata_kernel/issues/24. The user can specify `graphdir` in the config file, however. Here's some examples:

magic_globals
magic_plots


stata_kernel/kernel.py
```diff
@@ -53,9 +57,15 @@ def __init__(self, *args, **kwargs):
...
         config.read(os.path.expanduser('~/.stata_kernel.conf'))
+
         self.execution_mode = config['stata_kernel']['execution_mode']
         self.stata_path = config['stata_kernel']['stata_path']
 
+        try:
+            self.graphdir = config['stata_kernel']['graphdir']
+        except:
+            self.graphdir = tempfile.mkdtemp()
+
```
kylebarron: I updated this in dev: 
https://github.com/kylebarron/stata_kernel/blob/0450471d04f2a30a5d4ff9e10659ee811e409a15/stata_kernel/kernel.py#L37-L38

`StataSession` creates it if it doesn't exist:
https://github.com/kylebarron/stata_kernel/blob/0450471d04f2a30a5d4ff9e10659ee811e409a15/stata_kernel/stata_session.py#L36-L37

stata_kernel/kernel.py
```diff
@@ -116,13 +126,82 @@ def do_execute(
...
+        quit_early['status'] = 'ok'
+        quit_early['payload'] = []
+        quit_early['user_expressions'] = {}
+
+        if magic['name'] in stata_magics['global_magics']:
+            obj      = self.do("macro dir")
+            res      = obj.get('res').rstrip()
+            sglobals = magic['regex'].findall(res)
+            lens     = 0
+            pglobals = []
```
kylebarron: I haven't added a `Contributing.md` yet, but can you install [editorconfig](https://editorconfig.org/) and [yapf](https://github.com/google/yapf)? They'll use the project-level configuration files and make all code consistent, and won't make you touch any of your other projects.

stata_kernel/kernel.py
```diff
@@ -116,13 +126,82 @@ def do_execute(
...
+            r'gr(a|ap|aph)?',
+            r'tw(o|ow|owa|oway)?',
+            r'sc(a|at|att|atte|atter)?',
+            r'hist(ogram)?',
+            r'line',
+            r'kdensity',
+            r'lowess',
+            r'lpoly',
+            r'tsline'
+        ]
```
kylebarron: I discovered that there's actually a _lot_ of graph keywords. See `help graph_other`.

A simple list is fine but eventually I'll want to at least copy and paste the other graph keywords into the list.

stata_kernel/kernel.py
```diff
@@ -142,20 +221,28 @@ def do_execute(
         if silent:
             return return_obj
 
+        # TODO: Add a Hydrogen flag? REPL does not have this limitation
```
kylebarron: Yes, I'll change this back. It was just temporary to see stuff working!

stata_kernel/poor_mans_magic.py
```diff
@@ -0,0 +1,130 @@
...
+            #     if code.strip() != '':
+            #         magic['name']   = ''
+            #         magic['status'] = -1
+            #         print("Magic %{0} must be called by itself.".format(_magic))
+
+            else:
+                print("Uknown magic %{0}.".format(_magic))
+
+    elif code.strip().startswith("?"):
+        code = "help " + code
```
kylebarron: :+1:

stata_kernel/poor_mans_magic.py
```diff
@@ -0,0 +1,130 @@
...
+            # elif _magic in stata_magics['exit_magics']:
+            #     magic['name']    = _magic
+            #     magic['restart'] = (_magic == 'restart')
+            #     if code.strip() != '':
+            #         magic['name']   = ''
+            #         magic['status'] = -1
+            #         print("Magic %{0} must be called by itself.".format(_magic))
+
+            else:
+                print("Uknown magic %{0}.".format(_magic))
```
kylebarron: 'Unknown'

stata_kernel/poor_mans_magic.py
```diff
@@ -0,0 +1,130 @@
+import argparse
+import regex
+
+magic_regex = regex.compile(
+    r'\A%(?<magic>.+?)(?<code>\s+.*)?\Z',
```
kylebarron: `magic` and `code` are named capture groups? That you fill in using `stata_magics`?

mcaceresb: `magic` captures the word afer % if % is the first character in the code block. `code` captures all the things after. `stata_magics` is a dictionary that contains all the magics implemented. I wasn's sure what a nice way to implent magics was, but that is what I came uo with. Hance the `hack-ish` caveat.

stata_kernel/kernel.py
```diff
@@ -535,19 +620,44 @@ def get_graph_timestamp(self, name):
...
+            return [img]
+        else:
+            return []
+
+    def _makedirs_try(self, path_try):
+        try:
+            os.makedirs(path_try)
+        except OSError:
+            if not os.path.isdir(path_try):
+                raise
```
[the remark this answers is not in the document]
kylebarron: Also apparently `pathlib2` is a drop in replacement backported to 2.7 anyways.

kylebarron:
Also I meant to ask if you had an idea for the best way to learn plot status. Currently, if there's any graph keyword, after the block I check the timestamp on the graphs in memory. The problem is that the timestamps only exist to the minute resolution, and if the user makes two unnamed graphs in the same block, the second would overwrite the former anyways.

So I'm thinking of adding `graph export` after every line that I believe makes a graph.

stata_kernel/kernel.py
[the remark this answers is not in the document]
mcaceresb: Cool. Didn't know Path was available to recursively make dirs.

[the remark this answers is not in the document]
kylebarron: Yeah you can add `Path().mkdir(parents=True, exist_ok=True)`. It's great.

kylebarron:
I want to try and get the current dev branch decently stable enough to go into `master`, and then I'll check out your PR. Is there a way for me to edit your PR directly? Or can you merge it onto a different branch, that I could accept without it going straight to master?

This is the first time I've ever used Git or Github with someone else!

mcaceresb:
There is a general issue with the current approach that if a user saves a bunch of graphs the code would plot all of them instead of all the most recent ones.

AFAIK there is no reliable way in Stata to capture all the graphs, just the most recent ones. You also get the issue that a graph command can be embedded in a program, local, global, or loop.

Even if you use a hack like my `%plot` magic I wouldn't know how to reliably capture all the plots output by a block.

kylebarron:
> You also get the issue that a graph command can be embedded in a program, local, global, or loop.

Right. I'd momentarily forgotten that. I tokenize for blocks, so I could automatically add the `graph export` after all graph keywords that _aren't_ inside a block.

Even if it's not automatic, or not perfect automatically, I might try to think of a nice way to expose manual graph fetching through the UI.

mcaceresb:
When you merge to master git should automagically flag conflicts.

You can make edits to the PR directly.

Not sure how to make this a PR for dev, tho. I imagine it must be possible.

kylebarron:
Sounds good. I'll try to implement at least a basic graph exporting function that works with `dev` tomorrow. Then I'll merge to master and check out your pr.

Let me know if you come across any issues if you test out `dev`.

mcaceresb:
>  I could automatically add the graph export after all graph keywords that aren't inside a block.

Doesn't the code do that already? Or does it only do it for the last one even within a block?

I think printing all non-program/loop/etc. embedded graphs is fine (and for youse you can just print the last one). I believe this is how %matplotlib --inline does it?

kylebarron:
No it doesn't. `master` checks for keywords, and if any of the code matches the keywords, _then_ it checks the timestamps of the graphs in memory, and exports them if they're later than the timestamps the kernel has a record of.

mcaceresb:
@kylebarron Should I merge my branch with dev or just wait until you put dev in master?

kylebarron:
Just wait till I put dev in master. Your stuff is more of a new feature. I found a few more bugs this morning #52

mcaceresb:
@kylebarron Fixing this. It seems that the new plot checking mechanism has some issues. Will try to fix/work around:

error_graph

kylebarron:
Yes I know. I was going to fix that today. Currently for the console method, it inserts graph export always, and not just not in blocks. That should be quick for me to fix I think.

kylebarron:
image
I think that extra if clause is all you need, but I'm getting other errors now, so I need to test more.

kylebarron:
image

mcaceresb:
Just merged `upstream/master` and now I get:

```
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/mauricio/bulk/lib/stata_kernel/stata_kernel/__main__.py", line 4, in <module>
    IPKernelApp.launch_instance(kernel_class=StataKernel)
  File "/usr/lib/python3.6/site-packages/traitlets/config/application.py", line 657, in launch_instance
    app.initialize(argv)
  File "<decorator-gen-123>", line 2, in initialize
  File "/usr/lib/python3.6/site-packages/traitlets/config/application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)
  File "/usr/lib/python3.6/site-packages/ipykernel/kernelapp.py", line 465, in initialize
    self.init_kernel()
  File "/usr/lib/python3.6/site-packages/ipykernel/kernelapp.py", line 376, in init_kernel
    user_ns=self.user_ns,
  File "/usr/lib/python3.6/site-packages/traitlets/config/configurable.py", line 412, in instance
    inst = cls(*args, **kwargs)
  File "/home/mauricio/bulk/lib/stata_kernel/stata_kernel/kernel.py", line 23, in __init__
    self.stata = StataSession()
  File "/home/mauricio/bulk/lib/stata_kernel/stata_kernel/stata_session.py", line 90, in __init__
    self.stata.do(text)
AttributeError: 'StataSession' object has no attribute 'stata'
```

kylebarron:
oops

mcaceresb:
@kylebarron Note this embeds the timer into each `do_*` command to minimize the timer bias (ow it's up to 0.1 seconds). This means they return `(rc, timer)` instead of just `rc`.

This adds `%time [--profile]` magic as requested in https://github.com/kylebarron/stata_kernel/issues/51 (the switch benchmarks each individual line); %timeit would actually a bit more convoluted; I might actually make that a wrapper for `benchmark` or hold off on it for now.

magic_time

It also fixes `svg` image size and allows the user to set the default image size via `%plot --set`

magic_plots_after

This can be reset via `%plot --scale 1 --set`.

stata_kernel/stata_magics.py
```diff
@@ -0,0 +1,314 @@
+import argparse
+import regex
```
kylebarron: What's the benefit of using `regex` over the standard `re`? From what I can see it's backwards compatible with 2.7?

mcaceresb: I started using `regex` for named capture groups and recursive lookahead; now I'm just more used to it. Not sure if you can get either in `re`, but the specific syntax of my regexes would not port 1:1 to `re`.

mcaceresb: (PS: I'm not doing anything recursive, but I am doing named capture groups, which you can apparently get with a slightly different syntax; will port)

kylebarron: I wasn't necessarily saying I didn't want to use `regex`, I just didn't know. And I thought I was pushing my changes to my own branch 😕

mcaceresb: Jaja. Ok. It's working with `re` now anyway.

kylebarron:
I haven't dived into the code yet but it doesn't appear to be working for me on Mac (either automation or console):
image

But the sleep did happen in Stata on both.

mcaceresb:
Interesting... If you type `%foo` does it print anything?

kylebarron:
It appears to work when I use Jupyter console but not when I use Jupyter notebook. I don't know why:
image

kylebarron:
image

kylebarron:
But usually that would be sent straight to stata, so it's definitely upgraded. I don't see anything printed to console:
image

mcaceresb:
Mmm... I reckon something must happen with `stdout` bc I am just using `print`

kylebarron:
Is [your example](https://github.com/kylebarron/stata_kernel/pull/49#issuecomment-412310302) in the console??
How do you get plots to show up?

mcaceresb:
`jupyter qtconsole`

kylebarron:
Wow qtconsole is very cool.

In qtconsole I can't replicate your `%plot command`.

image

mcaceresb:
It's `--set`; `set` by itself is a Stata command.

stata_kernel/stata_magics.py
```diff
@@ -0,0 +1,297 @@
...
+            required=False)
+        self.timeit.add_argument(
+            '-n',
+            dest='n',
+            type=int,
+            metavar='N',
+            default=None,
+            help="Execute statement N times per loop.",
+            required=False)
+
```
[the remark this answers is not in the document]
mcaceresb: You can type `%magic -h` or `%magic --help`; this is added by default by `argparse`. It might make sense to make them named, tho.

stata_kernel/stata_magics.py
```diff
@@ -0,0 +1,297 @@
...
+            type=int,
+            metavar='height',
+            default=400,
+            help="Plot height",
+            required=False)
+        self.plot.add_argument(
+            '--set',
+            dest='set',
+            action='store_true',
+            help="Permanently set plot width and height.",
```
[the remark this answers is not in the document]
mcaceresb: Within the session.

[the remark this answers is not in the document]
kylebarron: Awesome. I could see a case for making a `[magics]` section of `~/.stata_kernel.conf` as well, but not sure if that's necessary.

stata_kernel/stata_magics.py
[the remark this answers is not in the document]
kylebarron: I didn't know. Sounds good!

kylebarron:
Btw with Automation the timing isn't as accurate. 
image

stata_kernel/stata_magics.py
```diff
@@ -0,0 +1,297 @@
...
+
+        return code
+
+    def post(self):
+        if self.timeit in [1, 2]:
+            total, l = self.time_profile.pop()
+            print("Wall time (seconds): {0:.2f}".format(total))
+
+            if (len(self.time_profile) > 0) and (self.timeit == 2):
+                tlens = 0
```
kylebarron: This is assigned to but never used.

kylebarron:
Oh you're using `print()` to return the magic message... That's why it's not working. You have to use [Jupyter-specific messaging functions](https://jupyter-client.readthedocs.io/en/stable/messaging.html#display-data).

mcaceresb:
This is now working on notebooks:

magic_nb_globals

magic_nb_time

stata_kernel/stata_magics.py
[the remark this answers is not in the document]
mcaceresb: This is not yet working on notebooks, tho =....=

stata_kernel/stata_magics.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+class MagicParsers():
+    def __init__(self):
+        self.plot = argparse.ArgumentParser()
+        self.plot.add_argument(
+            'code', nargs='*', type=str, metavar='CODE', help="Code to run")
+        self.plot.add_argument(
+            '--scale',
+            dest='scale',
```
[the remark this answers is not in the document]
mcaceresb: I think `-h` conflicts with the built-in `--help`

kylebarron:
Thanks. A couple last questions
- Whenever I get around to implementing autocompletions, I'm going to want to have a continuously- and automatically-updated dictionary in the class of locals, globals, variable names, (return values?), etc. That way the kernel can call the `do_complete()` function quickly without having to proxy the Stata session again. I'm fine merging this implementation, but just noting that for the future.
- I'm not sure what parts of the repo are possible to write tests for, and I'm not even sure I want to start writing tests yet, because things are still quite up in the air, but that's another thing to be thinking about. I currently have some tests for comments edge cases with the lexer, but that's it.
- I'll add some documentation for the magics to the website.

Anything else?

mcaceresb:
Help is working on notebooks as well.

magic_nb_help

mcaceresb:
@kylebarron 

> Whenever I get around to implementing autocompletions, I'm going to want to have a continuously- and automatically-updated dictionary in the class of locals, globals, variable names, (return values?), etc. That way the kernel can call the do_complete() function quickly without having to proxy the Stata session again. I'm fine merging this implementation, but just noting that for the future.

The issue is what to do as those things change, right? You can trigger `$<tab>` but then the user sets a given variable to "" and what do you do? A cache sounds like a good idea but not 100% how to keep it in sync.

> I'm not sure what parts of the repo are possible to write tests for, and I'm not even sure I want to start writing tests yet, because things are still quite up in the air, but that's another thing to be thinking about. I currently have some tests for comments edge cases with the lexer, but that's it.

This is kind of a GUI in many ways. Not sure how people test that...

stata_kernel/stata_magics.py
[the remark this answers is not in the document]
kylebarron: Ah yes of course.

stata_kernel/stata_magics.py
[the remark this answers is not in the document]
mcaceresb: This is now working on the notebooks as well. Not very happy with how I did it but it seems to work just fine.

kylebarron:
Well I don't see why I can't just do a `macro dir` after each `do_execute`. But this is a discussion for another thread.
image

mcaceresb:
@kylebarron Ah. I thought you wanted to avoid calling `macro dir` or `ds` post each execute. In that case yes, it shouldn't be too hard. `%globals` and `%locals` parse that already.

kylebarron:
It appears that `macro dir` is plenty fast 😛 
image

mcaceresb:
@kylebarron I intentionally truncated the timer digits to 2 because of slight inaccuracies from the 3-4th digit onward. `set rmsg on` seems to do it as well so I figured I was good.

kylebarron:
Did you add any new packages not in the standard library?

mcaceresb:
I don't think so; this is what I added:

```
from xml.dom import minidom
from timeit import default_timer
import argparse
```

I thought `xml`, `timeit`, and `argparse` were standard.

kylebarron:
Great, just wanted to double check

PR #63, opened 2018-08-12
Fixes SVG image aspect ratio

### LINKED ISSUE #62: Extra whitespace on image
@mcaceresb 
With the new PR, there's extra whitespace around images that are sent back to Jupyter. Compare

before:
image

after:

image

mcaceresb:
Fix the viewbox in SVG images when setting width and height so programs can set the correct aspect ratio. It also modifies the error regex to `r'\r\nr\((\d+)\);($|\r\n)'` because at least on my system Stata doesn't send the second line break; the string just ends.

- [X] closes #62 
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/stata_kernel/stata_magics.py
+++ b/stata_kernel/stata_magics.py
@@ -15,7 +15,6 @@ class StataParser(argparse.ArgumentParser):
         sys.exit(1)
 
     def error(self, msg):
-        print("debug1")
         print_kernel('error: %s\n' % msg, self.kernel)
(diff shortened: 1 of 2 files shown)


mcaceresb:
Also thinking to change the default in `%locals` and `%globals` to `--truncate` and instead allow users to specify `--verbose`; otherwise it might litter their screens if they have too many things defined...

kylebarron:
Switching to `--verbose` sounds fine.

PR #64, opened 2018-08-12
Support #delimit;

### LINKED ISSUE #37: Support #delimit ;
Basically have two sections of the lexer. One for ; and one for cr

kylebarron:
- [x] closes #37 
- [x] tests added / passed (will be done)

This PR 
- Updates the lexer to understand ;
- Tries to support edge cases according to [Statalist](https://www.statalist.org/forums/forum/general-stata-discussion/general/1448244-understanding-stata-s-comment-hierarchy). Tests forthcoming. The only thing it seemed I had to change was to prevent `\n` from ending the `*` comment, and to make it end with `;` instead.
- Inside `CodeManager`, if it sees there are chunks that use `#delimit;`, it removes the text newlines, substitutes `\n` for `;` and then runs it through the lexer once more to get blocks, i.e. `foreach` and `program`.

Issues still:
- [x] Sending just `#delimit ;` crashes stata because I apparently still send something?
- [x] Store #delimit state in the kernel class. This way someone can use #delimit; in one block and not need to copy it to every block.

### DIFF
--- a/stata_kernel/code_manager.py
+++ b/stata_kernel/code_manager.py
@@ -13,6 +13,13 @@ class CodeManager(object):
         self.tokens = self.tokenize_code(code)
         self.tokens_nocomments = self.remove_comments()
 
+        self.ends_sc = str(self.tokens_nocomments[-1][0]) in ['Token.Keyword.Namespace', 'Token.Keyword.Reserved']
+
+        self.has_sc_delimits = 'Token.Keyword.Namespace' in [str(x[0]) for x in self.tokens_nocomments]
+        if self.has_sc_delimits:
+            self.adjust_for_semicolons()
+
+
     def tokenize_code(self, code):
         """Tokenize input code using custom lexer
 
@@ -45,6 +52,23 @@ class CodeManager(object):
         """
         return [
             x for x in self.tokens if not str(x[0]).startswith('Token.Comment')]
+
+    def adjust_for_semicolons(self):
+        # Remove any \n with label Token.Keyword.Namespace
+        # These are embedded newlines inside #delimit ; blocks
+
+        tokens = [x for x in self.tokens_nocomments if not ((str(x[0]) == 'Token.Keyword.Namespace') and (x[1] == '\n'))]
+
+        # Change the ; delimiters to \n
+        tokens = [('Newline delimiter', '\n') if (str(x[0]) == 'Token.Keyword.Reserved') and x[1] == ';' else x for x in tokens ]
+
+        # and then join all text together
+        text = ''.join([x[1] for x in tokens])
+
+        # Then run it through the tokenizer again once more to get blocks
+        self.tokens_nocomments = self.tokenize_code(text)
+
+        self.get_chunks()
 
     def get_chunks(self):
         """Get valid, executable chunks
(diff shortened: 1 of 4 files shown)


kylebarron:
The lexer now passes all the tests on your Statalist post @mcaceresb

kylebarron:
@mcaceresb Can you test this?

image
image

kylebarron:
This has been working throughout my tests, so I'm gonna merge and we can deal with any followup issues in new issues.

mcaceresb:
Note that the actual rules are totally crazy like; something like

```
r"^\s*#\s*delimit(\s+(?!cr)|[^\s\w\.])"
r"^\s*#\s*delimit\s+cr"
```

Try it in a stata session:

```
#   delimit)cr just...why?
 # delimit cr)again, why?
```

Not 100% that you need to support this, but it's a fun little quirk.

mcaceresb:
You should def print the `delimiter is now {;|cr}` message, like stata does.

kylebarron:
Wow that's really crazy.
Yes you're right.

mcaceresb:
Some problems:

```stata
#delimit;
if ( 1 ) {;  // hangs
    disp "hi";
    };
```

```
#delimit cr
{
    disp "hi"
    #delimit;
    disp "bye";
    };
```

Gives
```
Unknown #command
bye; invalid name
r(198);
r(198);
```

```
#delimit cr
{
    disp "hi"
    #delimit;
    disp "
    b
    y
    e
    ";
    };
```

Gives 
```
hi
Unknown #command
```

And then
```
disp "hi" // hangs
disp "hi"; // also hangs
```

And gives
````
  File "/usr/lib/python3.6/site-packages/ipykernel/kernelbase.py", line 233, in dispatch_shell
    handler(stream, idents, msg)
  File "/usr/lib/python3.6/site-packages/ipykernel/kernelbase.py", line 399, in execute_request
    user_expressions, allow_stdin)
  File "/home/mauricio/bulk/lib/stata_kernel/stata_kernel/kernel.py", line 63, in do_execute
    rc, imgs, res = self.stata.do(cm.get_chunks(), self.magics)
  File "/home/mauricio/bulk/lib/stata_kernel/stata_kernel/stata_session.py", line 257, in do
    return rc, imgs, self.clean_log_console(log, new_syn_chunks)
  File "/home/mauricio/bulk/lib/stata_kernel/stata_kernel/stata_session.py", line 518, in clean_log_console
    assert log_line[0] == code_line
AssertionError
```

kylebarron:
Oh god. I never knew there could be so many edge cases. I should've waited for you earlier.

Lets take these one block at a time:
1. That's a `do_is_complete` error. I'll look into that.
2. That's a lexer issue. I currently only check for #delimit in the global scope and not inside a block.
3. Same as 2
4. That's a log file parsing (or in this case pexpect return) issue.  I think my log parsing is too strict and thus too fickle. I'm going to try to redo parts of that tonight.

Otherwise it's not crazy to impose a restriction on _some_ of these edge cases.

kylebarron:
1. should be fixed now https://github.com/kylebarron/stata_kernel/pull/67

kylebarron:
It's heartening that it appears Stata preprocesses #delimit;, so this still works:
image

That means it should be possible to improve the #delimit lexing, but I just need to more properly set it up into two complete passes of the input string.

kylebarron:
Let's move the discussion to the open issue.

PR #67, opened 2018-08-12
Some #delimit fixes

kylebarron:
This fixes:

- (Not a #delimit fix) Prevents completion when last token is a multiline comment
image
- Adds a magic named `%delimit` that prints the current delmiter
- Logs to the screen when the delimiter changes

### DIFF
--- a/stata_kernel/code_manager.py
+++ b/stata_kernel/code_manager.py
@@ -20,10 +20,10 @@ class CodeManager(object):
 
         self.has_sc_delimits = 'Token.Keyword.Namespace' in [
             str(x[0]) for x in self.tokens_nocomments]
-        self.is_complete = self._is_complete()
-
         if self.has_sc_delimits:
             self.adjust_for_semicolons()
+
+        self.is_complete = self._is_complete()
 
     def tokenize_code(self, code):
         """Tokenize input code using custom lexer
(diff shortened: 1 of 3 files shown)


kylebarron:
See this branch for progress:
https://github.com/kylebarron/stata_kernel/tree/dev-two-lexers

Do you think performance would suffer by lexing twice?

Also I'm kinda just considering writing my own stack, which would be a learning experience for me.

mcaceresb:
Two passes, one for comments one for parsing, is basically the conclusion I came to when trying to write a regex-based snippet to parse stata code.

> Do you think performance would suffer by lexing twice?

Probably not. That type of stuff will make performance suffer for pasting large text blocks. In those cases it seems acceptable for the parser to take its time.

For syntax highlighting this would be a big no no bc it would slow down text exiting, but that is not a concern here.

> Also I'm kinda just considering writing my own stack, which would be a learning experience for me.

Not sure what you mean.

kylebarron:
- This now uses two separate lexers.
- It seems to solve the first two issues, while the last issues are log-file parsing related
- It also removes the code that prevented both images and text from being sent (I can't reproduce my issues in Hydrogen).

I'm editing the tests I wrote so that they pass this. @mcaceresb would you like to test it out?

kylebarron:
Move this discussion to #70

Issue #48, opened 2018-08-10
Various ways the kernel is tripped because of changing prompts


mcaceresb:
#### Code Sample

```stata
disp "Sup?", _request(what_is_up)

pause on
program foo
    pause
    disp "bar"
end
foo

`local'

$global
```
#### Problem description

Each of the 4 examples above messes up the kernel because the prompt changes.

#### Package Version

```
Name: stata-kernel
Version: 1.1.0
Summary: A Jupyter kernel for Stata. Works with Windows, macOS, and Linux. Preserves program state.
Home-page: https://github.com/kylebarron/stata_kernel
Author: Kyle Barron
Author-email: barronk@mit.edu
License: GPLv3
Location: /usr/lib/python3.6/site-packages
Requires: jupyter-client, IPython, ipykernel, python-dateutil, pexpect
Required-by: 
```

kylebarron:
Can you try the `dev` branch?

mcaceresb:
Cool beans. It still gets tripped, tho. Btw is the code structure in `dev` stable? I'm meaning to give mata support another go.

kylebarron:
Actually ``` `local'``` and `$global` might be hard to fix. Because there's no way for me to know that it's invalid input.

mcaceresb:
Can you give me an example where using

```
re.compile(r'\r\n\. \Z', flags=re.MULTILINE)
```

to detect the end of Stata output fails? At a glance that would fix:

```
`local'
$global
qui do file
```

Then `\s\. \Z` seems to capture the first two cases; it can print something like "interactive output is not supported" or "incomplete input; aborting execution" or similar. The docstring notes "false positives" but I don't really know how to generate them.

mcaceresb:
This just happened to me accidentally: `!` enters a bash prompt for me. I can technically use bash inside stata inside bash inside stata... While it's tempting to redirect "!" and "shell" through python, it won't resolve globals and locals correctly. Maybe pipe the statement through `disp` and then use `subprocess`?

kylebarron:
`\Z`, like `$` in the regex would do nothing:
https://pexpect.readthedocs.io/en/stable/overview.html#find-the-end-of-line-cr-lf-conventions

mcaceresb:
> Pexpect reads from the child one character at a time

I'm not entirely see how that makes sense. Since passing the compiled regex seems to work on my system, can you humor me and find an example where it fails so I can let it go as "what works for me is not cross platform"?

[...]

PR #69, opened 2018-08-13
[Draft] Mata support

### LINKED ISSUE #26: Support Mata

mcaceresb:
Rudimentary mata support. **not** ready to merge. Wanted to get input on the TODOs.

- [ ] closes #26 
- [ ] tests added / passed
- [ ] whatsnew entry

Features:
- `mata` enters mata, `end` exits.
- `*` de-references a pointer and is not a comment.
- `program` and `input` get sent instead of prompting a continuation.
- `while` and `for` loops work as expected and prompt continuation.
- `%status` magic prints delimit and mata status.

TODO:
- [x] Mata-aware completions once completions are merged to master
- [x] `#delimit ;` is not currently working.
- [x] `if` is not currently working.
- [x] Code gets out of sync on continuation error
- [x] If there are no superfluous lines the code hangs (as does the regular kernel)

### DIFF
--- a/stata_kernel/code_manager.py
+++ b/stata_kernel/code_manager.py
@@ -8,18 +8,43 @@ class CodeManager(object):
     """Class to deal with text before sending to Stata
     """
 
-    def __init__(self, code, semicolon_delimit=False):
+    def __init__(self, code, semicolon_delimit=False, mata_mode=False):
+
+        # mata goes first since it obeys #delimit!
         if semicolon_delimit:
             code = '#delimit ;\n' + code
+            if mata_mode:
+                code = 'mata;\n' + code
+        elif mata_mode:
+            code = 'mata\n' + code
+
+        print('debugz0', code)
         self.input = code
         self.tokens = self.tokenize_code(code)
+        print('debugz1', self.tokens)
         self.tokens_nocomments = self.remove_comments()
+        print('debugz2', self.tokens_nocomments)
 
         self.ends_sc = str(self.tokens_nocomments[-1][0]) in [
             'Token.Keyword.Namespace', 'Token.Keyword.Reserved']
 
+        self.ends_mata = False
+        if len(self.tokens_nocomments) > 1:
+            token, chunk = self.tokens_nocomments[-2]
+            self.ends_mata = str(token) == 'Token.Keyword.Reserved'
+            print('debugz3.1', token, chunk)
+
+        print('debugz3', self.ends_sc, self.ends_mata, self.tokens_nocomments)
+
         self.has_sc_delimits = 'Token.Keyword.Namespace' in [
             str(x[0]) for x in self.tokens_nocomments]
+        self.has_mata_mode = 'Token.Other' in [
+            str(x[0]) for x in self.tokens_nocomments]
+        print('debugz4', self.has_sc_delimits, self.has_mata_mode)
+
+        if mata_mode:
+            self.tokens_nocomments = self.tokens_nocomments[1:]
+
         self.is_complete = self._is_complete()
 
         if self.has_sc_delimits:
--- a/stata_kernel/code_manager.py
+++ b/stata_kernel/code_manager.py
@@ -11,19 +11,22 @@ class CodeManager(object):
 
     def __init__(self, code, semicolon_delimit=False, mata_mode=False):
 
-        # mata goes first since it obeys #delimit!
+        # mata goes second since it obeys #delimit!
         code = re.sub(r'\r\n', r'\n', code)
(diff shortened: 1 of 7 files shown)


stata_kernel/code_manager.py
```diff
@@ -8,18 +8,43 @@ class CodeManager(object):
...
+                code = 'mata;\n' + code
+        elif mata_mode:
+            code = 'mata\n' + code
+
+        # print('debugz0', code)
         self.input = code
         self.tokens = self.tokenize_code(code)
+        # print('debugz1', self.tokens)
         self.tokens_nocomments = self.remove_comments()
+        # print('debugz2', self.tokens_nocomments)
```
kylebarron: The init to code_manager may be changed a bit. See 
https://github.com/kylebarron/stata_kernel/tree/dev-two-lexers

mcaceresb: Should I merge dev-two-lexers and continue from there?

kylebarron: I'd say so. I had to refactor a bit due to splitting into two lexers.

mcaceresb: Sometimes I feel like git is magic. Anyway, I merged in the dev-two-lexers branch. Same problems as before, tho. Hopefully with the new paradigm they will be easier to solve.

kylebarron: Which same problems?

mcaceresb: Mainly lack of `#delimit ;` support in mata and `if` trips it up (because in mata, if by itself does not finish execution. It wants an "else" statement).

kylebarron: Sounds like we need a Mata lexer as well...

mcaceresb: No, no. I think I can do it with the right nesting scheme.

mcaceresb: More generally, there are actually several instances where a line can be sent to mata, and the reply from stata can just be a line continuation prompt...

kylebarron: That sounds tough. Any way to force a dot prompt?

mcaceresb:
Well, that didn't go well. Will solve the conflicts at some point in the week.

kylebarron:
You might want to pull from master to see if that solves the travis build

mcaceresb:
Regardless, there is a problem here in that I am using `\Z' to break mata line continuations left open by the last mata command... However, per our discussion in #48, it is not always reliable.

kylebarron:
Well we should keep testing. But I'm worried there will be more cases that break down using only a single `\r\n\. `. The other thing you can try is passing a list. 
```
child.expect(['\r\n\r\n\. ', '\r\n\. ', pexpect.EOF])
```
The return value of this is the index of the list that was matched. This still runs into the issue of matching a `. ` in the results (maybe test out `'\r\n\. \Z'`) as the second value. And maybe if the result is `1` we wait a little to see if there's any text in the stream after? I'm really not sure.

stata_kernel/stata_lexer.py
```diff
@@ -46,17 +44,33 @@ class StataLexer(RegexLexer):
         'mata': [
             include('strings'),
             (r'^[^\n]*?\{', Token.MatchingBracket.Other, 'block'),
-            (r'^[^\n]*?\(', Token.MatchingBracket.Other, 'paren'),
-            (r'[\r\n][^\r\n\S]*end(?=(\s|[^\s\w\.]).*?$|$)', Token.Aborted, '#pop'),
+            (r'^[^\r\n]*?\(', Token.MatchingBracket.Paren, 'paren'),
+            (r'[\r\n][^\r\n\S]*end(?=(\s|[^\s\w\.]).*?$|$)', Token.Mata.Close, '#pop'),
```
kylebarron: How do you get arbitrary names?? I wouldn't be using `Token.MatchingBracket.Other` if I knew how to create my own Tokens!

mcaceresb: I didn't know you could use arbitrary names until earlier today. I kept getting errors so I looked it up. It turns out you can just name them `Token.WhateverTheHellYouWant` and it will take it.

kylebarron: Amazing. I may submit a PR to change the names, because `Token.MatchingBracket.Other` should be named like `Token.Block`

mcaceresb: I think at one point I use `startswith.('Token.MatchingBracketOther')`. I can just change it to `Token.Block.Paren` and it's all the same.

PR #70, opened 2018-08-13
Various fixes. Supercedes #67

kylebarron:
- [ ] closes #xxxx
- [x] tests added / passed

New features:

- This now uses two separate lexers. It updates the code_manager accordingly.
- It seems to solve the first two issues, while the last issues are log-file parsing related
- It also removes the code that prevented both images and text from being sent (I can't reproduce my issues in Hydrogen).
- It adds `stripall=False` and `stripnl=False` to the instantiations of the lexer classes. I believe this should give more accurate `is_complete` messages if you type `///` and then two newlines.
- Previous tests were updated and many more tests were added, especially with regards to `is_complete`. These should be easy to test. There's currently one failing test.

### DIFF
--- a/stata_kernel/code_manager.py
+++ b/stata_kernel/code_manager.py
@@ -4,6 +4,7 @@ from pygments import lex
 from .stata_lexer import StataLexer
 from .stata_lexer import CommentAndDelimitLexer
 
+
 class CodeManager(object):
     """Class to deal with text before sending to Stata
     """
@@ -58,8 +59,7 @@ class CodeManager(object):
             (List[Tuple[Token, str]]):
                 list of non-comment tokens
         """
-        return [
-            x for x in tokens if not str(x[0]).startswith('Token.Comment')]
+        return [x for x in tokens if not str(x[0]).startswith('Token.Comment')]
 
     def convert_delimiter(self, tokens):
         """If parts of tokens are `;`-delimited, convert to `\\n`-delimited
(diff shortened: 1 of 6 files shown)


mcaceresb:
```stata
#delimit ;
disp
"
a
2
b
"
;
```

prints
```
 a  2 b c
```

That is, line breaks are replaced with a space. However, in the kernel this prints
```
a 2bc
```

kylebarron:
Fixed 

image

kylebarron:
The reason for the below chunk failing is that usually I had `"` string blocks set to close at every newline. When not in #delimit; blocks, You can run `di "hi` and it'll work just fine. 

**However** this brings up a possibly-important point. I currently have the line continuation indent set to 4 spaces. But apparently those spaces are sent as part of the code. I can't accurately remove them because a user could do a backspace to delete them. Should I remove the continuation indent entirely?

```
#delimit cr
{
    disp "hi"
    #delimit;
    disp "
    b
    y
    e
    ";
    };
```
image

mcaceresb:
I vote for removing the indent (at least in delimit).

kylebarron:
Do you think I should remove it always or just in #delimit blocks?

kylebarron:
@mcaceresb 
Do you understand why this doesn't work?
```
disp "Line start"
// This is ignored, and does not give error
"; Line end" ;
```
But this does?
```
disp "Line start"
 // This is ignored, and does not give error
"; Line end" ;
```

I don't understand why the first wouldn't work. It just needs a space at the beginning of the line?

kylebarron:
So when a `//` comment is in the middle of a block, it has to have at least one space after the newline?

I'm not sure I can get that into a regex...

kylebarron:
image

Ok I'm going to merge this and we can work out some minor details in future pull requests. I want to have this version being pip-installable tomorrow morning for the other RA's ;)

PR #73, opened 2018-08-13
%magic completions, %completions magic, completions fixes

### LINKED ISSUE #92: Local macro completion not working
@mcaceresb I'm wondering if the new PR yesterday messed this up. Can you try from master at your leisure?

(The suggestion in Atom is from the language-stata package, there should be another one with a Hydrogen icon with the same suggestion).

image
image

mcaceresb:
- `%x<tab>` completions (only match if at start of code)
- `%set plot` sets plot settings for the current session.
- `%set --reset _all` resets all settings to their defaults.
- `%set --permanently` modifies the system config file.
- `%set x<tab>` matches available settings.
- `%globals` and `%locals` now show [regexp] instead of [code] in their help
- `%locals`, `%globals` by default truncate output (this should make it nicer
  if the user has several long values saved). To print the full contents
  the user can pass option --verbose.

- [X] closes #92, #93 
- [ ] tests added / passed
- [ ] whatsnew entry


kylebarron:
> %x<tab completions (only match if at start of code)

What does that do?

mcaceresb:
Completions for magics.

kylebarron:
I'm guessing this has to be changed to work with current master?

kylebarron:
Let me know when you think it's ready to be merged

mcaceresb:
Whenever you like. I'm tackling html help atm (should be a short commit). After that I'll move on to mata.

stata_kernel/code_manager.py
```diff
@@ -218,22 +218,33 @@ def get_text(self, config):
...
-        graph_fmt = config.get('graph_format')
-        graph_scale = config.get('graph_scale')
+        over_fmt = config.overrides['plot']['format']
+        over_scale = config.overrides['plot']['scale']
+        over_width = config.overrides['plot']['width']
+
+        graph_fmt = over_fmt if over_fmt else config.get('graph_format')
+        graph_scale = over_scale if over_scale else config.get('graph_scale')
+        graph_width = over_width if over_width else 600
+        graph_height = config.overrides['plot']['height']
```
kylebarron: So these settings allow for manual aspect ratio sizing, right? I assume there's no issue if a user hasn't set a default in their `conf` file?

kylebarron: So `config.overrides` will give what if it doesn't exist?

stata_kernel/code_manager.py
```diff
@@ -218,22 +218,33 @@ def get_text(self, config):
...
         cache_dir = config.get('cache_dir')
         if graph_scale is None:
             graph_scale = 1
             config.set('graph_scale', '1', permanent=True)
         else:
             graph_scale = float(graph_scale)
 
+        dim_str = " width({})".format(int(graph_width * graph_scale))
+        if graph_height:
+            dim_str += " height({})".format(int(graph_height * graph_scale))
```
kylebarron: :+1: I ran into an issue with them being floats.

stata_kernel/kernel.py
```diff
@@ -26,8 +26,10 @@ def __init__(self, *args, **kwargs):
 
         # Can't name this `self.config`. Conflicts with a Jupyter attribute
         self.conf = Config()
-        self.magics = StataMagics()
+        self.graphs = {}
+        self.graph_formats = ["svg", "png", "pdf", "tif"]
```
kylebarron: Unless you're also changing `send_image`, this should be just `svg` and `png`. `pdf` support doesn't really exist... in Jupyter Notebook the user is presented with a link to open the PDF in a new tab. I don't see any reason why `tif` would improve on either `svg` or `png`; it's like 1-2 orders of magnitude larger too.

Do you use `self.graphs` again?

stata_kernel/kernel.py
```diff
@@ -168,15 +170,22 @@ def do_is_complete(self, code):
...
+                'status': 'ok',
+                'cursor_start': pos,
+                'cursor_end': cursor_pos,
+                'matches': self.completions.get(chunk, env, rcomp)}
+        else:
+            return {
+                'status': 'ok',
+                'cursor_start': cursor_pos,
+                'cursor_end': cursor_pos,
+                'matches': []}
```
kylebarron: So this is a setting for turning completions off entirely? At least in Hydrogen this can also be turned off client side. Does the kernel need a setting for it? In Jupyter notebook and the console you only see it when you use `tab`, and there's not really another use for `tab` in those environments...

stata_kernel/config.py
```diff
@@ -24,6 +25,17 @@ def __init__(self):
...
+        # where applicable.
+        self.overrides = {
+            'plot': {
+                'format': None,
+                'scale': None,
+                'width': None,
+                'height': None
+            }
+        }
+
```
kylebarron: Can you just make this part of `Config.env`? Specifically, the `config.get()` function has a `permanent` option, and if that's `False`, it won't be written to the configuration. So in my mind, the `%set` magic would be a wrapper around `config.set(key, val)`. Just use `plot_format`, `plot_scale`, `plot_width`, and `plot_height`. Also, the current format configuration variable is `graph_format`, not `plot_format`. So that will get confusing. I'm happy to change that to `plot_format`, but there are several places in the code where we need to make that change.

stata_kernel/stata_magics.py
```diff
@@ -58,10 +57,51 @@ def __init__(self, kernel):
...
+        #######################################################################
+        #                                                                     #
+        #                             %set magic                              #
+        #                                                                     #
+        #######################################################################
+
+        self.set = StataParser(prog='%set', kernel=kernel)
+        # self.set.add_argument(
+        #     '--permanently', dest='perm', action='store_true',
+        #     help="Store settings permanently", required=False)
```
kylebarron: Why is this commented out?

mcaceresb: Ah, this was meant to have a TODO attached; the idea is that if the `--permanently` is passed, this gets set in the `.conf` file, but I wasn't sure what was the best place to put it on (specially since I didn't put a ton of stuff under `%set`

stata_kernel/config.py
mcaceresb: Ok.

stata_kernel/kernel.py
mcaceresb: I can take it out. I think this was legacy from there I was concerned about it causing any lag. But even if the user has a ton of things defined I don't think the lag will be appreciable.

stata_kernel/kernel.py
mcaceresb: I thought `self.graphs` was from master. I don't use it. I'll take out `pdf` and `tif'

stata_kernel/code_manager.py
mcaceresb: The idea was to set it locally or to whatever is in `config`. I'll change it to work with `env`

stata_kernel/kernel.py
kylebarron: `self.graphs` was used a long time ago when I was trying to keep a manual record of what graphs were exported. I recently deleted it as I had forgotten about it.

stata_kernel/code_manager.py
kylebarron: :+1: I've been meaning to try to figure out how to write the configuration to file including comments, but I haven't done that yet. So as of now, writing the configuration to file deletes the initial explanatory comments written during the initial `install.py` script.

stata_kernel/kernel.py
kylebarron: Ah I see. If you find yourself bothered by lag we can put it back in.

mcaceresb:
Merging master actually reverted the change I had made that fixed the locals thing... Trying to remember what I did a few hours ago...

kylebarron:
You should be able to look at the commit before merging master?

mcaceresb:
It apparently didn't revert my change. So it's something else that re-broke it.

mcaceresb:
Figure it out. I was accounting for a results-cleaning issue that you fixed.

stata_kernel/completions.py
```diff
@@ -270,7 +266,7 @@ def get_suggestions(self, kernel):
 
             all_locals = """mata : invtokens(st_dir("local", "macro", "*")')"""
             res = '\r\n'.join(
-                re.split(r'[\r\n]{1,2}', self.quickdo(all_locals, kernel))[1:])
+                re.split(r'[\r\n]{1,2}', self.quickdo(all_locals, kernel)))
```
kylebarron: :+1:

stata_kernel/config.py
```diff
@@ -77,3 +69,10 @@ def raise_config_error(self, option):
         https://kylebarron.github.io/stata_kernel/user_guide/configuration/
         """.format(option)
         raise ValueError(dedent(msg))
+
+    def _remove_unsafe(self, key, permanent=False):
+        self.env.pop(key, None)
+        if permanent:
+            self.config.remove_option(option=key, section='stata_kernel')
+            with self.config_path.open('w') as f:
+                self.config.write(f)
```
kylebarron: What does this do?

mcaceresb: Removes a config option. Namely I use it to remove graph_width and graph_height, which in effect resets them to their defaults.

As the function name implies, it is unsafe in that it does not check for whether the config should be removed. Careful internal use only!

PR #80, opened 2018-08-16
WIP Return results asynchronously

kylebarron:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

Is on the same branch as the last two PRs. Those should be merged first. For now, this removes the magics from `do` while I get the base working again.

I have to figure out how to clean the log stream to only show results and not input. I'm not sure what the best way to do that is.

### DIFF
--- a/stata_kernel/stata_session.py
+++ b/stata_kernel/stata_session.py
@@ -167,8 +167,8 @@ class StataSession(object):
         Kwargs:
             magics (StataMagics):
                 magics.graphs:
-                    0 dooes not look for images
-                    1 looks after each line not in Token.MatchingBracket.Other
+                    0 does not look for images
+                    1 looks after each line not in Token.TextBlock
                     2 looks only after _all_ lines have executed
                 magics.timeit:
                     0 do not time it
@@ -228,7 +228,7 @@ class StataSession(object):
                     break
 
                 gr = re.search(graph_keywords, line[1]) and (graphs == 1)
-                if (str(line[0]) != 'Token.MatchingBracket.Other') and gr:
+                if (str(line[0]) != 'Token.TextBlock') and gr:
 
                     rc, img, sc = self.get_current_graph(
                         'console', magics.img_metadata)
@@ -278,7 +278,7 @@ class StataSession(object):
             for line in syn_chunks:
                 syn_chunk_counter += 1
                 new_syn_chunks.append(line)
-                if str(line[0]) == 'Token.MatchingBracket.Other':
+                if str(line[0]) == 'Token.TextBlock':
                     rc, timer = self.do_aut_async(line[1])
                 else:
                     rc, timer = self.do_aut_sync(line[1])
@@ -517,7 +517,7 @@ class StataSession(object):
 
         log_all = []
         for (Token, code_line), log_line in zip(syn_chunks, log):
-            if str(Token) != 'Token.MatchingBracket.Other':
+            if str(Token) != 'Token.TextBlock':
                 # Since I'm sending one line at a time, and since it's not a
                 # block, the first line should equal the text sent
                 # The assert is just a sanity check for now.
@@ -565,7 +565,7 @@ class StataSession(object):
         # DoCommandAsync
         syn_chunks_new = []
         for (Token, code_lines) in syn_chunks:
-            if str(Token) == 'Token.MatchingBracket.Other':
+            if str(Token) == 'Token.TextBlock':
                 syn_chunks_new.append((Token, 'cap noi ' + code_lines))
             else:
                 syn_chunks_new.append((Token, code_lines))
@@ -580,7 +580,7 @@ class StataSession(object):
         # this for loop, I set the first line to `inexact` matching.
         last_whitespace = ''
         for (Token, code_lines) in syn_chunks:
-            if str(Token) != 'Token.MatchingBracket.Other':
+            if str(Token) != 'Token.TextBlock':
                 # Means I sent it with DoCommand; there should be no leading
                 # spaces. Also means it should be a single line
                 all_code_lines.append(['exact', last_whitespace + code_lines])
--- a/stata_kernel/stata_session.py
+++ b/stata_kernel/stata_session.py
@@ -14,6 +14,7 @@ from configparser import ConfigParser
 if platform.system() == 'Windows':
     import win32com.client
     from win32api import WinExec
+    from pexpect import fdpexpect
 else:
     import pexpect
 
@@ -34,26 +35,6 @@ ansi_regex = r'\x1b(' \
              r'(\d;\dR))'
 ansi_escape = re.compile(ansi_regex, flags=re.IGNORECASE)
 
-graph_keywords = [
-    r'gr(a|ap|aph)?', r'tw(o|ow|owa|oway)?', r'sc(a|at|att|atte|atter)?',
-    r'line', r'hist(o|og|ogr|ogra|ogram)?', r'kdensity', r'lowess', r'lpoly',
-    r'tsr?line', r'symplot', r'quantile', r'qnorm', r'pnorm', r'qchi', r'pchi',
-    r'qqplot', r'gladder', r'qladder', r'rvfplot', r'avplot', r'avplots',
-    r'cprplot', r'acprplot', r'rvpplot', r'lvr2plot', r'ac', r'pac', r'pergram',
-    r'cumsp', r'xcorr', r'wntestb', r'estat\s+acplot', r'estat\s+aroots',
-    r'estat\s+sbcusum', r'fcast\s+graph', r'varstable', r'vecstable',
-    r'irf\s+graph', r'irf\s+ograph', r'irf\s+cgraph', r'xtline'
-    r'sts\s+graph', r'strate', r'ltable', r'stci', r'stphplot', r'stcoxkm',
-    r'estat phtest', r'stcurve', r'roctab', r'rocplot', r'roccomp',
-    r'rocregplot', r'lroc', r'lsens', r'biplot', r'irtgraph\s+icc',
-    r'irtgraph\s+tcc', r'irtgraph\s+iif', r'irtgraph\s+tif', r'biplot',
-    r'cluster dendrogram', r'screeplot', r'scoreplot', r'loadingplot',
-    r'procoverlay', r'cabiplot', r'caprojection', r'mcaplot', r'mcaprojection',
-    r'mdsconfig', r'mdsshepard', r'cusum', r'cchart', r'pchart', r'rchart',
-    r'xchart', r'shewhart', r'serrbar', r'marginsplot', r'bayesgraph',
-    r'tabodds', r'teffects\s+overlap', r'npgraph', r'grmap', r'pkexamine']
-graph_keywords = r'\b(' + '|'.join(graph_keywords) + r')\b'
-
 
 class StataSession(object):
     def __init__(self):
@@ -65,10 +46,9 @@ class StataSession(object):
         config = ConfigParser()
         config.read(Path('~/.stata_kernel.conf').expanduser())
 
-        cache_dir = config['stata_kernel'].get(
-            'cache_directory', '~/.stata_kernel_cache')
-        cache_dir = Path(cache_dir).expanduser()
-        cache_dir.mkdir(parents=True, exist_ok=True)
+        self.cache_dir = Path(config['stata_kernel'].get(
+            'cache_directory', '~/.stata_kernel_cache')).expanduser()
+        self.cache_dir.mkdir(parents=True, exist_ok=True)
         self.execution_mode = config['stata_kernel'].get('execution_mode')
 
         stata_path = config['stata_kernel'].get('stata_path')
@@ -85,7 +65,6 @@ class StataSession(object):
             'png': ' width({0}) height({1})'}
 
         self.stata_path = stata_path
-        self.cache_dir = cache_dir
         self.graph_format = config['stata_kernel'].get('graph_format', 'svg')
         self.graph_size = graph_export_size[self.graph_format]
         self.graph_cmd = 'qui graph export `"{0}/graph.{1}"\' , as({1}) replace'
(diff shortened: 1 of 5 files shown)


stata_kernel/stata_session.py
```diff
@@ -154,261 +136,103 @@ def init_console(self):
...
-        Args:
-            line (str): literal string ready to send to Stata
-        Returns:
-            (int): return code from Stata
-            (int): execution time
-        """
+        md5 = '`' + md5 + "'"
+        error_re = r'^r\((\d+)\);'
+        graph_notice = '// This is the stata kernel graph notice'
+        more = r'--more--'
```
mcaceresb: Shouldn't `more` start with `^`?

I also wonder if there is a way to implement a small (0.1s or something) timeout after matching `more` or `error_re` since a user can print `disp "--more--"` or "disp "r(198);". This wouldn't be an issue with the other stopping conditions, since a user wouldn't be able to print the MD5 hash or the graph notice (at least not without trying).

kylebarron: I added `^` to more. I think there are no spaces...

I think those timeouts are something to come back to later if they cause a problem. It's probably pretty rare for someone to want to display `--more--` or `r(198);`.

kylebarron: I'm having issues with not being able to stop more actually.

```py
if match_index == 3:
    self.child.send('q')
    break
```
This is sending `q` to the tty whenever it sees `--more--`, but in the TTY it isn't breaking.

image

mcaceresb: For more it wouldn't be too hard, right? Something like `expect(r"^(?!--more--\Z).+", timeout = 0.1)` and if it times out then there is nothing printed after `--more--`. The error would be trickier. Not sure what is a good idea for that.

mcaceresb: You need to send about 3 newlines/carriage returns because when you quit more it wants to give you `error 1`

kylebarron: Because then `--more--` wouldn't be caught by that... It would be caught by one of the other rules, like the end of line. Or do you mean I'd do that after I've already successfully expected for `^--more--`?

mcaceresb: Yeah. If you match more, then expect something that isn't more. If you don't get it then `--more--` was a stopping point. Otherwise it was printed by the user for some reason. That regex should match anything that isn't `^--more--$`

mcaceresb: Fun fact: When I re-size my terminal, the pagesize resets to whatever Stata now thinks is reasonable. I doubt this will ever be caught out by that (though sometimes it seems you need more than 3 CR to break `more`...

kylebarron: > You need to send about 3 newlines/carriage returns because when you quit more it wants to give you error 1

In my terminal, when I press `q`, the more breaks. Then I need to expect for 3 newlines, maybe, but I don't need to send three newlines. Here, sending q isn't doing anything.

mcaceresb:
Idea:

```
lineLength = n = response.find("\n")
codeRegex = r"^({})$".format(r'\n\. '.join([r'\n >'.join([line[i:i+n] for i in range(0, len(line), n)]) for line in codeLines]))
re.sub(codeRegex, '', response)
```

Might be too slow?

mcaceresb:
Actually that wouldn't work to find the line length, but you ned the idea: Find the line length and then use a regex to replace input code.

mcaceresb:
You can try to match lines of code im order. So instead of trying to match all the lines of code every time. try to match the first line if there is a prompt line r"\n\. "; if it maches go to codeLines.next() otherwise continue printing the stream.

kylebarron:
I'm getting an issue with using fdpexpect on Automation where with the default search buffer it hits the end of the file before catching the MD5. I'm wondering if this is because it's trying to read more bytes than the file descriptor is providing (https://github.com/pexpect/pexpect/pull/304#issuecomment-164174490). 

When I set `maxread=1`, it works fine, but that might slow down the reading a bit.

stata_kernel/stata_session.py
kylebarron: Note to self, apparently `ctrl-C` then `ctrl-D` also breaks a `more`. Test that out with expect.

kylebarron:
idea for removing code lines:
```py
# split text into lines
lines = text.split('\n')
# For the first line, try to match `. {lines[0][:75]}`, i.e. the first
# 75 characters of the first line. (75, or linesize - 5) is chosen so
# that it catches lines that are `  1. ` inside a program or for loop
#
# If it's a match, look at child.before to see how many characters were
# matched. If the line had more characters than were matched, take off
# the first 75 characters, prepend `> ` and try to match again.
# When the full line is matched, remove the first indexed object and
# repeat.
```

kylebarron:
@mcaceresb would you like to test now?

kylebarron:
Also works on Mac automation. Interestingly enough, it actually shows up in the terminal _before_ in the Stata GUI. Which would seem to say that Stata writes the results to the log file before showing them on the screen.

kylebarron:
It should work with long input lines that wrap beyond `c(linesize)`, but as of now, any extra long output will be displayed as shown in the Stata console. It would be very easy to create a magic that allows a user to widen the linesize, and I could decide to widen past 80 characters by default.

kylebarron:
I still have to
- [ ] Get the timer working again. @mcaceresb do you think I should use `rmsg` instead of doing it from python?
- [ ] Allow for setting graph size. This should be easy.
- [ ] Decide when to use `include`. I'm thinking of setting it to be any selection with at least one newline. Then if the first command raises an error, the second won't run. I imagine that most of the time that you're working interactively, you'll be typing one line at a time. So that way you'll still have low latency most of the time.
- [ ] Test on Windows and Linux. I don't expect any Windows issues. But I think I should probably pass Linux results through the ANSI escape code regex. I think you might currently see escape codes with the banner or from something like `!ls`.
- [ ] Update docstrings and documentation website.

kylebarron:
@mcaceresb You know how you were frustrated how pexpect took 50ms to send a line? That's actually an [imposed default of the package](https://pexpect.readthedocs.io/en/latest/api/pexpect.html#spawn-class).

> The delaybeforesend helps overcome a weird behavior that many users were experiencing. The typical problem was that a user would expect() a “Password:” prompt and then immediately call sendline() to send the password. The user would then see that their password was echoed back to them. Passwords don’t normally echo. The problem is caused by the fact that most applications print out the “Password” prompt and then turn off stdin echo, but if you send your password before the application turned off echo, then you get your password echoed. Normally this wouldn’t be a problem when interacting with a human at a real keyboard. If you introduce a slight delay just before writing then this seems to clear up the problem. This was such a common problem for many users that I decided that the default pexpect behavior should be to sleep just before writing to the child application. 1/20th of a second (50 ms) seems to be enough to clear up the problem. You can set delaybeforesend to None to return to the old behavior.

I set the delay to `None` for now to at least try it out. Granted it makes way less of a difference now that I'm only sending a single command to the child at a time instead of one for each line.

image
image

kylebarron:
I'm thinking of completely removing the `%plot` magic, since it's not really necessary anymore. Maybe providing an option within `%set` for graph size instead? (I need to remove the save warnings)
image

mcaceresb:
1. Can't you just `qui graph export`? Just be careful to do `os.path.isfile` if you're relying on that message somehow.

2. Removing plot is fine if the kernel always catches them, right? Then it makes more sense in `%set`. Is the kernel catching all of them because it always checks for plots or is it doing something else?

3. I don't think parsing `set rmsg` is a good idea. I mean, the user can set it off, for one. And the user might set it on and then you'd be stripping them when you shouldn't. Things like that. It seems like extra work for not all that much payoff. Using timer seems OK to me. I can put it back once async gets merged.

4. Are you relying on a set linesize? If the user sets it, would that be a problem?

kylebarron:
> 1. Can't you just `qui graph export`? Just be careful to do `os.path.isfile` if you're relying on that message somehow.

I could do `qui graph export`, but I am using that message. That also makes it nicer because the user can do `qui program_name` to prevent graphs showing. Adding `isfile` is a good idea, but probably unnecessary since that message only shows if there's a graph open:
image


> 2. Removing plot is fine if the kernel always catches them, right? Then it makes more sense in `%set`. Is the kernel catching all of them because it always checks for plots or is it doing something else?

Yeah I'm planning to move it to `%set`. The kernel catches all of them because it uses the `(file /Users/kyle/.stata_kernel_cache/Graph1.png written in PNG format)` message on a successful graph output.

> 3. I don't think parsing `set rmsg` is a good idea. I mean, the user can set it off, for one. And the user might set it on and then you'd be stripping them when you shouldn't. Things like that. It seems like extra work for not all that much payoff. Using timer seems OK to me. I can put it back once async gets merged.

Sounds good. Thanks.

> 4. Are you relying on a set linesize? If the user sets it, would that be a problem?

Only partially. I want to check if a line that starts with a period is a code line and not an output line. I can't just do `code_line in result_line` because if `code_line` is longer than `linesize`, it won't work. So I currently check if the first `linesize - 5` characters of `code_line` are in `result_line`. By default, I have `linesize` set to 80, but that would be very easy to add to `%set`. 

I could also require fewer of the code characters to be present in `result_line`. If I had it set to `65`, then if linesize were reduced to 70 without the kernel's knowledge, everything would still work. Alternatively, widening the `linesize` without the kernel's knowledge is not a problem.

kylebarron:
It might be helpful to remove `stata_kernel_graph_counter` from the list of globals that are suggested as completions and from `%globals`.

mcaceresb:
- it's not too hard to trip the current plot strategy. One problem seems like an OK fix (scanning for the keywords not just at the start), but I feel like the point of the `%plot` magic was that you wouldn't have to account for every possible edge case. Anyway, here are some ways to trip it:

```
sysuse auto
hist price
disp "Just did a graph of price"
```

Shows the histogram twice. Then
```
local note note("The graph shows a density")
local kden kdensity price
`kden', `note'
```

Shows the histogram after the first line, again after the second line, and nothing in the third line.

- I worry about the effect of `set linesize 40`; It doesn't really seem to be an issue, and I can't think of a quick way to trip it.

> It might be helpful to remove stata_kernel_graph_counter from the list of globals that are suggested as completions and from %globals.

I think the first is a good idea but I'm not so sure about the second... Maybe it's fine since `macro dir` would show it anyway?

kylebarron:
Yes you're right. Maybe only search the beginning of the string? That would probably be a happy medium

mcaceresb:
> Yes you're right. Maybe only search the beginning of the string? That would probably be a happy medium

I think your default of showing a plot if one was successfully exported is nice, given you can clear the previous plot. However, it seems like too much pain to deal with plots inside

- locals and globals
- do/include calls
- programs
- mata functions
- classes

If the user generates a bunch of plots through a convoluted call, I also think it's very understandable if only the last one gets displayed (which is already the default behavior of the GUI AFAIK).

kylebarron:
Haha why? You like really skinny graphs or really wide graphs?

mcaceresb:
> Haha why? You like really skinny graphs or really wide graphs?

Picky reason most users might not care about: Stata's default aspect ratio is 11:8 or something funky like that. AFAIK this doesn't change for different plots. The default in the kernel is 3:2, right? This seems better but it's not like every plot ever should be like that. My brain defaults to 4:3, but setting an image ratio to your computer's resolution ratio is quite reasonable as well (16:9, etc. might actually be more reasonable).

Actual decent reason to allow it: Stata can combine graphs. It may or not make sense to keep a given aspect ratio for the combined graphs (specially if they are not combined in a square fashion).

mcaceresb:
I just thought of another good one (at least for something I do a non-trivial amount): Stata's note wrapping for graph notes is terrible. I have definitely done the process of:

1. Split up note into chunks
2. Plot.
3. See if note overflows.
4. Restart from 1 if it overflows.
5. ???
6. Profit.

Allowing slight changes in graph width could help with that (making a plot slightly wider vs re-doing a note split).

kylebarron:
Ok fine those are good points. I'm leaning towards pushing that to another PR though. This is already _quite_ bloated.

mcaceresb:
Sure.

kylebarron:
Can you try `!ls` on linux and see if you see ansi escape codes?

mcaceresb:
```
!ls
HANGELOG.md  LICENSE       requirements_dev.txt  setup.py
docs          MANIFEST.in  requirements.txt      stata_kernel
fluff         README.md    setup.cfg             tests
[?1h=
```

kylebarron:
Ok thanks. `HANGELOG.md` isn't ideal...

mcaceresb:
It cuts stuff off.

```
!pwd
e/mauricio/bulk/lib/stata_kernel
[?1h=
```

kylebarron:
That's weird. It doesn't on Mac.
image

mcaceresb:
Newline issue?

```
!printf "\n"
[?1l>
[?1h=
```

mcaceresb:
```
!printf "\r"
[?1h=
```

kylebarron:
The escape codes are a Stata-terminal communication issue. The cut-off text is separate I think.

mcaceresb:
Try `!top`? Is nothing cut  off? (That will trip up the kernel, but I'm curious to know if this is 100% a Linux thing.)

kylebarron:
I'm unable to reproduce on NBER servers.
image

mcaceresb:
Try qtconsole.

mcaceresb:
Or a notebook

mcaceresb:
error_escape

kylebarron:
Ok well pull the latest commit and the escape codes should at least be gone

mcaceresb:
This seemed to fix all the things for me.

g2g; will try to add mata tonight. Will chunk all the magic, etc. commits so no commit is too bloated.

Looking pretty good tbh! It's def faster atm. Once I add mata I'll start trying this out as a daily driver. Exciting!

kylebarron:
Good to hear! Even the cut off at the beginning? Do you think passing text through the complicated ANSI_escape regex is slow enough to do it only on Linux?

It should be quite a bit faster!

mcaceresb:
Btw I do think there should be a section gathering all the warnings about how interactive commands don't work: more, pause, display, _request(), certain shell commands, etc.

kylebarron:
Yes, that's a valid point.

PR #88, opened 2018-08-20
Add %browse

kylebarron:
- [x] closes #87
- [x] tests added / passed

@mcaceresb it could probably be improved by adding `varlist`, and maybe `if`. I figure it would take you only a couple minutes to figure out the right `argparse` line? The problem with `if` is that I don't want to export the dataset, and I'm not sure if there's a way to _first_ apply the `if` and _then_ apply the `in`. Because using both of them returns nothing if there are no observations within `in` that also meet `if`.

PR #89, opened 2018-08-20
Add if/in/varlist to %browse

kylebarron:
Continuation of #88

### DIFF
--- a/stata_kernel/ado/_StataKernelExportDelim.ado
+++ b/stata_kernel/ado/_StataKernelExportDelim.ado
@@ -1,0 +1,30 @@
+* Written by Mauricio Caceres
+* August 19, 2018
+* version 0.1
+program _StataKernelExportDelim
+    syntax [varlist] [if/] [in/] using, [*]
+    if ( "`if'" != "" ) {
+        local stype = cond(`=_N' < maxlong(), "long", "double")
+        tempvar touse sumtouse
+        gen byte `touse' = `if'
+        gen `stype' `sumtouse' = sum(`touse')
+        local last = `=`sumtouse'[_N]'
+
+        gettoken f in: in, p(/)
+        gettoken s l:  in, p(/)
+
+        if ( "`l'" == "" ) {
+            local ifin (`sumtouse' == `f')
+        }
+        else {
+            local ifin (`sumtouse' >= `f') & (`sumtouse' <= `l') & `touse'
+        }
+    }
+    else if ( "`in'" != "" ) {
+        local ifin in `in'
+    }
+    else {
+        local ifin
+    }
+    export delimited `varlist' `using' `ifin', `options'
+end
--- a/stata_kernel/stata_magics.py
+++ b/stata_kernel/stata_magics.py
@@ -40,6 +40,14 @@ class MagicParsers():
             required=False)
 
         self.browse = StataParser(prog='%browse', kernel=kernel)
+        self.browse.add_argument(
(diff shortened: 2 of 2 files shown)


stata_kernel/ado/_StataKernelExportDelim.ado
```diff
@@ -0,0 +1,30 @@
...
+        tempvar touse sumtouse
+        gen byte `touse' = `if'
+        gen `stype' `sumtouse' = sum(`touse')
+        local last = `=`sumtouse'[_N]'
+
+        gettoken f in: in, p(/)
+        gettoken s l:  in, p(/)
+
+        if ( "`l'" == "" ) {
+            local ifin (`sumtouse' == `f')
```
mcaceresb: This was missing ``& `touse'`` at the end.

stata_kernel/ado/_StataKernelExportDelim.ado
```diff
@@ -0,0 +1,30 @@
...
+        }
+        else {
+            local ifin (`sumtouse' >= `f') & (`sumtouse' <= `l') & `touse'
+        }
+    }
+    else if ( "`in'" != "" ) {
+        local ifin in `in'
+    }
+    else {
+        local ifin
```
mcaceresb: I think this can default to `local ifin in 1 / 200`, right?

mcaceresb:
Here's a version that  falls back on `list`

```
program _StataKernelExportDelim
    syntax [varlist] [if/] [in/] using, [*]

    if ( !`=_N > 0' ) error 2000

    if ( "`in'" == "" ) local in 1/200

    if ( "`if'" != "" ) {
        local stype = cond(`=_N' < maxlong(), "long", "double")
        tempvar touse sumtouse index
        gen byte `touse' = `if'
        gen `stype' `sumtouse' = sum(`touse')
        gen `stype' `index' = _n
        local last = `=`sumtouse'[_N]'
        gettoken f in: in, p(/)
        gettoken s l:  in, p(/)
        if ( "`l'" == "" ) {
            local ifin (`sumtouse' == `f') & `touse'
        }
        else {
            local ifin (`sumtouse' >= `f') & (`sumtouse' <= `l') & `touse'
        }
    }
    else {
        local ifin in `in'
    }

    qui export delimited `index' `varlist' `using' `ifin', replace `options'
    list `varlist' `ifin'
end
```

Then you can model the `_browse` magic after head to get the Stata index and send `rest` as fallback. I think it's better to let Stata handle the syntax parsing in this case, and just spit out the Stata error message in `argparse` and show the Stata syntax in the help, but that might be a matter of taste, I suppose.

kylebarron:
Sure. The only thing I want to make sure is ok is if there aren't 200 observations in the dataset. Like with
```
sysuse auto
%browse
```
Will that automatically add `in 1/200`? Then Stata will give an index error...

mcaceresb:
```
if ( "`in'" == "" ) local in 1/`=min(200, _N)'
```

kylebarron:
ah brilliant!

kylebarron:
Even when `in` is provided though, you should still do ``in 1/`=min(`in', _N)``, because it's possible `if` could subset to a very small number of people, and then `in` wouldn't be able to do, say, `1/100`.

mcaceresb:
`if _n == 1 in 1/100` should not give an error in the program.

`in 1/100` if there are only 100 obs would give an error, but I think that, philosophically, that _should_ give an error, but it's your call.

Issue #13, opened 2018-08-06
Code introspection


kylebarron:

kylebarron:
A simple way to do this without scraping is just to run `man command` when code introspection is desired.

kylebarron:
See https://github.com/kylebarron/stata_kernel/pull/97#issuecomment-414513509

PR #97, opened 2018-08-20
%help magic displays HTML help where supported.

### LINKED ISSUE #86: Support rich output types
You can have [many types of output](https://nbviewer.jupyter.org/github/ipython/ipython/blob/master/examples/IPython%20Kernel/Rich%20Output.ipynb) embedded in a Jupyter Notebook.

- [LaTeX](https://nbviewer.jupyter.org/github/ipython/ipython/blob/master/examples/IPython%20Kernel/Rich%20Output.ipynb#LaTeX); sadly it seems as if MathJax doesn't support tabular, so this might actually not be too useful.
- [HTML](https://nbviewer.jupyter.org/github/ipython/ipython/blob/master/examples/IPython%20Kernel/Rich%20Output.ipynb#HTML); There probably aren't too many Stata commands that export HTML...

mcaceresb:
- [ ] tests added / passed
- [ ] whatsnew entry

Fixes #86 

Adds %help magic that displays HTML help if supported (tested in Atom and Notebook).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+beautifulsoup4>=4.6.3
 jupyter_client>=5.2.3
 IPython>=6.5.0
 ipykernel>=4.8.2
@@ -6,4 +7,3 @@ pillow>=5.2.0
 pexpect>=4.6.0
 pygments>=2.2.0
 regex>=2018.7.11
-bs4>=0.0.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ pillow>=5.2.0
 pexpect>=4.6.0
 pygments>=2.2.0
 regex>=2018.7.11
+bs4>=0.0.1
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -4,3 +4,4 @@ include requirements.txt
 include requirements_dev.txt
 include LICENSE
 include stata_kernel/ado/*.ado
+include stata_kernel/css/*.css
--- a/stata_kernel/css/_StataKernelHelpDefault.css
+++ b/stata_kernel/css/_StataKernelHelpDefault.css
@@ -1,0 +1,31 @@
+h2 {
+    font-family: Arial,Helvetica,Helv,sans-serif;
+    color: #000000;
+}
+
+pre {
+    margin: 10px;
+}
+
+table {
+    background-color: transparent;
+    border-color: transparent;
+    bgcolor: transparent;
+}
+
+tr {
+    background-color: transparent;
+    border-color: transparent;
+    bgcolor: transparent;
+}
+
+body {
+    background-color: transparent;
+    border-color: transparent;
+    margin: 0px;
+}
+
+div {
+    background-color: transparent;
+    border-color: transparent;
+}
--- a/stata_kernel/stata_magics.py
+++ b/stata_kernel/stata_magics.py
@@ -1,8 +1,13 @@
 import sys
 import re
+import urllib
 import pandas as pd
+
+from pathlib import Path
 from argparse import ArgumentParser
+from bs4 import BeautifulSoup as bs
 from .code_manager import CodeManager
+from pkg_resources import resource_filename
 
 
 # NOTE(mauricio): Figure  out if passing the kernel around is a problem...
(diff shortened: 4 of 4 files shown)


mcaceresb:
Note this adds BeautifulSoup4 as a dependency (bs4).

kylebarron:
I haven't been testing my code against `flake8` (I'd copied that from pandas), but I have been meaning to test code against yapf..

mcaceresb:
I have a plugin that yells at me when my code is not PEP8 compliant.

mcaceresb:
I found two "issues" with this:

1. Relative references don't work on atom (but they do seem to work on notebooks)
2. The font size seems small on Atom, but I wonder if that's just my system (this is also fine on notebooks).

Let me know if you find any problems. The background and font color look OK to me on Atom.

kylebarron:
Quite nice. Is it Hydrogen or is there an extra bar at the bottom?
image

I especially like that it shows you the underlining, which it didn't seem you could get from `man help`.

kylebarron:
image
better screenshot

kylebarron:
oooh the links are a very nice touch

image

kylebarron:
Though they aren't perfect; several of them link to inside the notebook:

http://localhost:8888/notebooks/Untitled1.ipynb?kernel_name=stata#weight
http://localhost:8888/notebooks/Untitled1.ipynb?kernel_name=stata#continuous_opts
http://localhost:8888/notebooks/Untitled1.ipynb?kernel_name=stata#discrete_opts
http://localhost:8888/notebooks/Untitled1.ipynb?kernel_name=stata#options

mcaceresb:
Several are supposed to link inside the notebook. They are meant to be relative.

But I just realized I might be over-assuming the way Stata does that.

requirements.txt
```diff
@@ -6,3 +6,4 @@ pillow>=5.2.0
 pexpect>=4.6.0
 pygments>=2.2.0
 regex>=2018.7.11
+bs4>=0.0.1
```
kylebarron: Is this better than just putting in `beautifulsoup4>=4.6.0`?

mcaceresb:
All the examples you listed work correctly for me in the notebook (i.e. they jump to the correct place in the help)

kylebarron:
Oh **wow**. I didn't even click on them to try that. I assumed they wouldn't work for some reason.

That extra line in my screenshot is an extra `<hr>`, and appears to be directly from the bottom of Stata's help page. I think it looks bad, and I might see how easy it is with beautifulsoup to just remove it.
image
image

kylebarron:
Thanks :grinning:

mcaceresb:
Ok, so the relative parsing should now only keep the relative link if it starts with 'help_command#reference'. e.g. 'histogram#options' would get a relative reference, but 'graph#examples' would not.

mcaceresb:
I can't get them to work on Atom, tho...

mcaceresb:
Actually, I have `beautifulsoup4` 4.6.1 as well. Mmm...Not sure how that works. Maybe both?

kylebarron:
If you do `pip show bs4`, it shows that the version is `0.0.1` and that it requires `beautifulsoup4`. 

The [documentation](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#installing-beautiful-soup) suggests installing `beautifulsoup4`, so that's probably preferable to put in the dependencies. I think if you install `beautifulsoup4` you can still run `from bs4 import beautifulsoup`

mcaceresb:
Sounds good.

kylebarron:
Separately, it would be nice to use this also for `do_inspect` responses. See #13  In Hydrogen this [shows up as a separate pane](https://nteract.gitbooks.io/hydrogen/docs/Usage/GettingStarted.html#code-introspection), which would be helpful when you want to see the help file next to your code without taking up the main screen.

The [`do_inspect` API](https://jupyter-client.readthedocs.io/en/stable/wrapperkernels.html#MyKernel.do_inspect) is nearly the same as the completions API, and the [response API](https://jupyter-client.readthedocs.io/en/stable/messaging.html#msging-inspection) is very close to a display_data message, but the difference is that here that dict is the return object, and not sent through `kernel.send_response`.

It should be possible to reuse this almost entirely. I did start [a branch](https://github.com/kylebarron/stata_kernel/tree/dev-do-inspect) to parse the input but it could definitely be improved.

PR #116, opened 2018-08-22
Mata support

### LINKED ISSUE #26: Support Mata

mcaceresb:
- [X] closes #26 
- [x] tests added / passed
- [ ] whatsnew entry

Features:
- `mata` enters mata, `end` exits.
- `*` de-references a pointer and is not a comment.
- `program` and `input` get sent instead of prompting a continuation.
- `while` and `for` loops work as expected and prompt continuation.
- `if` without `else` is evaluated correctly (a note is shown to the
   user noting the input was incomplete)
- `#delimit ;` support
- Supports nested parentheses `()`
- Supports brackets `{}`
- `%status` magic prints delimit and mata status.

This also features mata-aware completions: mata of objects are
suggested, locals are triggered by `` ` `` and globals by `$`. However,
`st_` functions also trigger completions:

    st_local("<tab>
    st_matrix("<tab>
    st_numscalar("<tab>
    st_strscalar("<tab>
    st_data(., "<tab>

would show completions for

    locals
    matrices
    scalars
    scalars
    varlist

respectively.


kylebarron:
I won't really be able to test this at all since I've never used mata.

mcaceresb:
Is there a better time to learn? I think not! FD

kylebarron:
Haha I really don't think there's much value in me learning Mata. 1) If I'm most likely leaving the economics world in a year, I may never use Stata again. 2) I use Stata because it's what Joe uses and is best for continuity. But if I have to use something other than Stata, why use Mata when I can use Python? If I assume that the next RA won't already know Mata, then there's really no usefulness to doing any of my work in Mata instead of Python.

mcaceresb:
Mata can interact with Stata in a way python cannot; but if the choice is pure mata vs pure python of course you ought to go with python. I started using mata to skirt some Stata annoyances but I've done several things in mata that end up being quite useful. Several parts of `gtools` for example use mata for small things. The power sim code uses mata for performance reasons. Whenever I do MLE in Stata I use the mata interface instead of the Stata interface bc it's much easier to use. Etc.

kylebarron:
Have you had any issues with this branch in the last two weeks?

mcaceresb:
No, it's been good, but no rush to merge or anything. Though I find mata rather useful, I do have to agree that "interactive mata support" isn't necessarily a top priority issue.

kylebarron:
Ok. I'll come back to this. I labeled with "Milestone 1.5" the issues I want to solve first.

mcaceresb:
Note to self: Look into rope completion for mata classes.

@kylebarron On a different note: Any thoughts on keyword completion for Stata commands? Also, is there any merit to even trying to parse syntax statements for Stata programs or user programs?

kylebarron:
rope completion?

> Any thoughts on keyword completion for Stata commands?

I personally don't have a need because `language_stata` already does this.

> Also, is there any merit to even trying to parse syntax statements for Stata programs or user programs?

What's the point of this?

mcaceresb:
When you type `object.<tab>` in python it gives you methods for the object; I thought that was called rope completion. Maybe that's just in my head?

>> Also, is there any merit to even trying to parse syntax statements for Stata programs or user programs?

> What's the point of this?

If you type

```
cmd anything, <tab>
```

It would autocomplete the options for that command; I was asking if you thought that was an idea worth pursuing. It might be too much of a hassle, tho.

kylebarron:
Maybe it is. I didn't know.

That would be cool, but it sounds incredibly difficult to get right. You'd suggest parsing the help files?

mcaceresb:
After some googling, it seems that's what some vim plugins call it, so maybe it's not so universal.

Yes. That was my idea. Parse from `program` until the _first_ `^\s*syntax` or the first `^\s*end`, then get all the keywords after the comma.

One issue is that it can never be perfect because Stata has no common syntax for programs, so a program could catch syntax in various parts and parse it later, etc.

kylebarron:
> Yes. That was my idea. Parse from program until the first ^\s*syntax or the first ^\s*end, then get all the keywords after the comma.

That sounds like you're parsing the source files... but you don't see the source files of stata commands?

kylebarron:
Apparently you can see the `.ado` files of (some ?) Stata commands. But whenever you have `*` you wouldn't know what `kwargs` are accepted.

kylebarron:
@mcaceresb 
Does this enter mata and exit mata in every code execution? Or does it start an interactive mata session and then end it? Is there a benefit to having a persistent mata session vs opening and closing it each time?

mcaceresb:
Interactive session. I reckon it's functionally the same, in the end. I'm sure I did whichever was easier to code at the time.

kylebarron:
So if you start mata, define an object, end mata, and go into mata again, that object will still be there, right?

mcaceresb:
Yeah.

kylebarron:
Anyways, I'm trying to sort out these merge conflicts

mcaceresb:
I can do it, unless you feel strongly you should? Let me know and I'll get started.

kylebarron:
If you want to, you're welcome. I haven't done much yet.

mcaceresb:
Where do you export graphs now?

kylebarron:
Where? 
https://github.com/kylebarron/stata_kernel/blob/b804ce96bc02cbd5616e6fc5fb70029e35c9d23f/stata_kernel/code_manager.py#L247-L265

So 
```
noi gr export {cache_dir_str}/graph${gph_cnt}.{graph_fmt}
```

mcaceresb:
Cool.

mcaceresb:
- Solved merge conflicts
- `quickdo()` wraps call in `stata()` if within mata
- `quickdo()` in `kernel.py` strips mata leading character.
- Mata completions for built-ins whenever 2+ characters are typed.
- Completions now show correct vector name (no size).

kylebarron:
I don't think the prepending is still necessary

mcaceresb:
It was actually causing problems, it seems.

On a tangentially related note, should paths only get suggested if there is a slash?

kylebarron:
There could be a path in the current directory. Maybe I should limit paths to just do files or something?

mcaceresb:
Right, but then you would use `./`

kylebarron:
I think most stata users wouldn't use "./". They'd just use the name.

mcaceresb:
Interesting point. Counter: typing `reg <tab>` and seeing the paths as suggestions is confusing. Typing `randomCommand <tab>` and seeing paths may or not be confusing, but `randomCommand ./<tab>` and seeing paths is def expected.

kylebarron:
Yeah. It could definitely be improved

kylebarron:
The `_mata_builtins.py` might go better in the `pygments/` subdirectory.

kylebarron:
Apparently IPython suggests file paths everywhere:
image

mcaceresb:
> Apparently IPython suggests file paths everywhere:

Maybe do variables first and paths later? I think ipython sorts the entire list;  maybe variables, then paths, then files?

kylebarron:
IPython sorts the entire list.
image

Note `lib/` and `local/` in the middle.

I think that's easiest to implement.

kylebarron:
@mcaceresb 	Also do you have any idea about last night's issue?

kylebarron:
image

1. It would be good to have a message saying "Now in Mata", or, even better, a line:
```
------------------------------------------------- mata (type end to exit) -----
```

    Apparently, that is shown for `mata:` but not `mata` 
image




2. The `#delimit ;` command doesn't work inside mata. Do you consider that a bug?

mcaceresb:
> I think that's easiest to implement.

Counter: ipython does have _some_ context awareness (import completes only modules by default; functions suggest their arguments, objects + `.` suggest their methods, etc.). Since we're not doing too much of that in Stata, grouping variables, folders, and files in the suggestions list might be sensible.

You mean the linesize stuff? Not really. It's weird... There's no reason for Stata to cut off the output like that.... I mean, maybe when it opens a graphics "window" it changes the linesize temporarily for some reason?

mcaceresb:
> #delimit ;
> Apparently, that is shown for `mata:` but not `mata` 

I vaguely recall trying to debug both of these and it being surprisingly hard. Will try again. If you do `#delimit ;` before `mata` it will work.

Try %status instead of `%mata`.

mcaceresb:
Ok, the reason is that `mata:` and `mata` actually have different behavior on error (one exits mata on error, the other does not), so they are parsed differently. Will see if I can normalize the parsing while keeping the correct behavior.

kylebarron:
Anyways, no time pressure on this.

mcaceresb:
Actually, the culprit is [here](https://github.com/kylebarron/stata_kernel/blob/master/stata_kernel/stata_session.py#L382-L384). Since `mata` is in the output (bc it says "mata (type end to exit)") it `res` is returned as blank.

mcaceresb:
Scratch that; it's the inverse of what I said.

mcaceresb:
Actually, the issue was that the `mata` line was parsed with the mata prompt, but the first line still has the Stata prompt.

mcaceresb:
The `#delimit ;` thing I won't debug any time soon, tho. That'd involve messing with the lexer and I don't think debugging that will be as easy.

mcaceresb:
Where?

kylebarron:
Check out https://github.com/kylebarron/stata_kernel/blob/master/tests/test_stata_lexer.py

kylebarron:
`python -m pytest tests/`

kylebarron:
The small note I'd say... As of now the `linesize` is set to 255 to minimize the likelihood that some paths are long enough that the graph doesn't show. But this means that Stata gives very wide dash lines when entering Mata. If it were up to me, I'd hide the Mata line that Stata gives and just send our own Mata line of width 80.

image

mcaceresb:
I'm not entirely sure how I'd do that when mata is opened or closed in the middle of a code block. I'd have to be watching for mata opening and closing in pexpect, but I'm hesitant to fiddle with that.

kylebarron:
Yeah don't bother

mcaceresb:
Btw `desc` and similar functions also have the linesize display issue.
screenshot_20181101_123120

kylebarron:
Well the way to fix that is to figure out how to deal with line continuations in results. Otherwise stuff like the messages `graph exported to /file/path...` can be wrapped over multiple lines, and then they aren't caught. 

So do you think it would be ok to say that every line that starts with `> ` is a line continuation of the previous line? There's no way for me to always know what the `linesize` is, because it could've been changed within a program.

kylebarron:
@mcaceresb 	You think this is good to merge?

mcaceresb:
I think so. I've been working off this branch all week and last week. I'm happy with it, and I did add some tests.

Issue #103, opened 2018-08-21
more woes


mcaceresb:
#### Code Sample

```stata
more
```
#### Problem description

`more` crashes the kernel.

#### Expected Output

No crash.

#### Debugging log

From `tail`

```
. more
`addec426932e71323700afa1911f8f1c'
more
        
. addec426932e71323700afa1911f8f1c'
command addec426932e71323700afa1911f8f1c is unrecognized
r(199);
```

#### Package Version

```
Name: stata-kernel
Version: 1.3.1
Summary: A Jupyter kernel for Stata. Works with Windows, macOS, and Linux. Preserves program state.
Home-page: https://github.com/kylebarron/stata_kernel
Author: Kyle Barron
Author-email: barronk@mit.edu
License: GPLv3
Location: /usr/lib/python3.6/site-packages
Requires: beautifulsoup4, jupyter-client, IPython, ipykernel, pandas, pillow, pexpect, pygments, regex
Required-by: 
```

mcaceresb:
If you leave more on, it should be set to

```
local usermore: set more
// stuff
set more `usermore'
```

kylebarron:
My idea is to set more by default when starting the kernel. Then if someone has a `more` option at the top of their script, that will override the kernel.

I'm still struggling to figure out the ideal way to send a break to the console... Working with the raw Stata console itself:
- Doing <kbd>Ctrl</kbd>-<kbd>D</kbd> just scrolls down `pagesize`
- Doing <kbd>Ctrl</kbd>-<kbd>C</kbd> on its own does nothing
- Doing <kbd>Ctrl</kbd>-<kbd>D</kbd> then <kbd>Ctrl</kbd>-<kbd>C</kbd> scrolls once and then does visibly nothing
- Doing <kbd>Ctrl</kbd>-<kbd>C</kbd> then <kbd>Ctrl</kbd>-<kbd>D</kbd> successfully sends a break, but it appears to remove the first character of the next line. It appears that <kbd>Ctrl</kbd>-<kbd>D</kbd> by itself removes the following character when not in a break:
![stata_ctrl_d](https://user-images.githubusercontent.com/15164633/44410293-155a7e00-a532-11e8-97fb-4c32791db5be.gif)
- Pressing <kbd>q</kbd> works in the Stata console, but doesn't seem to send a break with `child.send('q')` or `child.sendline('q')`, the latter of which also sends an end of line character.

@mcaceresb Any ideas?

[...]

PR #117, opened 2018-08-22
[WIP] Code introspection.

### LINKED ISSUE #13: Code introspection

mcaceresb:
- [X] closes #13
- [ ] tests added / passed
- [ ] whatsnew entry

One issue is that the kernel currently eats up sequential newlines, even if surrounded by non-blank output. This causes the formatting to be awkward. The help format would be fixed if the whitespace issue is fixed.

This displays plain text help even in atom. To fetch HTML help would slow-down introspection in other environments, but it's really easy to enable that; I recycled the code from `magic_help` so you can enable it if you want in `do_inspect`.

Last, this is also affected by the `--more--` bug in #103 because `help keyword` usually triggers more. The introspection would work correctly if that bug is fixed. The solution for now is to set more off before triggering introspection.


kylebarron:
Tomorrow I'm taking a vacation until Sept 4. Once I get back I'll look at these PRs more closely.

mcaceresb:
Bon voyage!

kylebarron:
> One issue is that the kernel currently eats up sequential newlines, even if surrounded by non-blank output. This causes the formatting to be awkward. The help format would be fixed if the whitespace issue is fixed.

This should be fixed now.

> This displays plain text help even in atom. To fetch HTML help would slow-down introspection in other environments, but it's really easy to enable that; I recycled the code from magic_help so you can enable it if you want in do_inspect.

How does introspection work in qtconsole? It's just `?word`? I'm assuming the initial `?` isn't included in what Jupyter sends to the kernel? 

I personally think it's fine to enable HTML all the time.
This could be behind a setting, but these HTML files should be really small. Even `help syntax` is only 30KB, so should take a fraction of a second. If we implement basic ANSI-code formatting, having the HTML would help for that too.

> The solution for now is to set more off before triggering introspection.

I feel that it's probably best to always set more off.

mcaceresb:
`(` triggers it.

My concern is not "response time given slow connection" rather "DNS lookup timeout time given the user is not connected to the internet." I'm good with the former bc, as you say, the files are small. The latter can take several seconds.

For something that is triggered as you type, it couod be somewhat frustrating.

kylebarron:
Well introspection isn't triggered as you type in any frontend I don't think.

It's triggered by `?` in console plus an extra command option in Hydrogen I think

mcaceresb:
`?` in the console just brings up the help, no?

It's triggered as you type in qtconsole by `(`; I don't think the HTML is working for me in hydrogen, but there is a slight lag in qtconsole so it's definitely trying to download something.

kylebarron:
> ? in the console just brings up the help, no?

I thought that was _because_ of introspection. Maybe not

mcaceresb:
[Here](https://github.com/kylebarron/stata_kernel/blob/master/stata_kernel/stata_magics.py#L232-L233)

PR #146, opened 2018-09-07
Newlines fixes

### LINKED ISSUE #111: Withhold whitespace-only lines until a non-whitespace result line is found
Would improve display of:
image

### LINKED ISSUE #144: Missing newlines
@mcaceresb 
> Is this issue (#111) what causes 
> ```
> disp _n(10) "hi" _n(10) "bye"
> ```
> To eat all the newlines?

The issue here is that I expect for `\n`, so there are no `\n`s in the result lines, and then I only display a result using `if res`. So '^\n' gets expected to '' and the `bool` of that is `False`.

Of course displaying those newlines can be desirable, but I don't want to display newlines for lines I sent to Stata. So I think I'll have to change the code here a bit:
https://github.com/kylebarron/stata_kernel/blob/2438b515d1ef314cc3fd6165d9d575b030ffdc3b/stata_kernel/stata_session.py#L256-L266

kylebarron:
- [x] closes #144, closes #111
- [ ] tests added / passed
- [ ] whatsnew entry

- Keep empty newlines of output
- Don't return whitespace-only lines until a text character is found. This means that running `di _n(3)` will give no result. The reason for this is that it looks better in Hydrogen to have a check mark to represent the fact that a command finished with no material output, rather than seeing empty whitespace lines. For example:
image

### DIFF
--- a/stata_kernel/stata_session.py
+++ b/stata_kernel/stata_session.py
@@ -258,9 +258,13 @@ class StataSession():
                 if res == '':
                     res += '\n'
                 if res:
-                    res_list.append(res)
-                if display and res:
                     res = ansi_escape.sub('', res)
+                    res_list.append(res)
+                if not ''.join(res_list).strip():
+                    continue
+                if display and res:
+                    if not ''.join(res_list[:-1]).strip():
+                        res = ''.join(res_list[:-1]) + res
                     self.kernel.send_response(
                         self.kernel.iopub_socket, 'stream', {
                             'text': res,
--- a/stata_kernel/stata_session.py
+++ b/stata_kernel/stata_session.py
@@ -255,7 +255,7 @@ class StataSession():
                 break
             if match_index == 4:
                 code_lines, res = self.clean_log_eol(child, code_lines, res)
-                if res == '':
+                if res is not None:
                     res += '\n'
                 if res:
                     res = ansi_escape.sub('', res)
@@ -278,7 +278,7 @@ class StataSession():
         child.expect('\r?\n')
 
         # Remove line continuation markers in output returned internally
-        res = '\n'.join(res_list)
+        res = ''.join(res_list)
         res = res.replace('\n> ', '')
 
         return rc, res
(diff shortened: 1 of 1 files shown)


kylebarron:
@mcaceresb Can you test this?

kylebarron:
image
There we go

kylebarron:
But now there's an extra newline at the end of each commmand:
image
vs
image

kylebarron:
What do you mean?

mcaceresb:
This seems to work for me (up to trailing newlines)

- add `res_disp = ''` before the loop
- After

```
                if not ''.join(res_list).strip():
                    continue
```

Add

```
                if not res_disp.strip():
                    continue
```

- Instead of passing `res`, pass `res_disp` and set `res_disp = ''` when you do.

mcaceresb:
I don't really know how to submit PRs to existing PRs. Sorry for the verbosity.

kylebarron:
Just write as a code segment the entire block...

kylebarron:
Like as markdown in the issue

mcaceresb:
```python
        match_index = -1
        res_list = []
        res_disp = ''
        rc = 0
        while match_index != 0:
            match_index = child.expect(expect_list, timeout=None)
            res = child.before
            if match_index == 0:
                break
            if match_index == 1:
                rc = int(child.match.group(1))
                if display:
                    self.kernel.send_response(
                        self.kernel.iopub_socket, 'stream', {
                            'text': 'r({});\n'.format(child.match.group(1)),
                            'name': 'stderr'})
                continue
            if match_index == 2:
                if display:
                    self.kernel.send_image(child.match.group(1))
            if match_index == 3:
                self.send_break(child=child)
                break
            if match_index == 4:
                code_lines, res = self.clean_log_eol(child, code_lines, res)
                if res is None:
                    continue
                res += '\n'
                res = ansi_escape.sub('', res)
                res_disp += res
                res_list.append(res)
                if not ''.join(res_list).strip():
                    continue
                if not res_disp.strip():
                    continue
                if display:
                    self.kernel.send_response(
                        self.kernel.iopub_socket, 'stream', {
                            'text': res_disp,
                            'name': 'stdout'})
                    res_disp = ''
                continue
            if match_index == 5:
                sleep(0.05)
```

kylebarron:
Like before that only works for me for leading whitespace and not trailing whitespace:
image

kylebarron:
Sweet

PR #183, opened 2018-09-16
PDF Graph redundancy

### LINKED ISSUE #175: Send (svg or png) AND (pdf) image representations to frontend
## Overview 
The quagmire of image formats is such:

### SVG

**Pro**
- Vector format
- Low file size
- Viewable in all rich frontends

**Con**
- Can't be used with LaTeX without converting
- Can't be created with earlier versions of Stata

### PDF

**Pro**
- Possible to create on any platform and any version of Stata
- Vector format
- Low file size
- Can be exported to LaTeX without conversion

**Con**
- Not viewable in _any_ frontend

### PNG

**Pro**
- Possible to create in earlier versions of Stata
- Viewable in all rich frontends

**Con**
- Not vector image format
- Can't be created on Linux

## Proposal

I propose that `stata_kernel` always returns both the (SVG or PNG) _and_ the PDF representation of all images back to the frontend.

1. When multiple mimetypes of the data are sent to Jupyter Notebook, it chooses which to show based on an internal preference hierarchy of mimetypes. But crucially, it doesn't discard alternate data. So when I send a `display_data` object that has _both_ `image/svg+xml` _and_ `application/pdf` mimetype objects, both are stored inside the `.ipynb` file. This means that when I export the Notebook to an HTML file, it uses the SVG version, but when I export to a PDF, it uses the PDF version of the image and doesn't touch the SVG data.

  This means that on a Linux server without inkscape, I was able to view the SVG image in the Notebook and in the exported HTML, but I was still able to compile a LaTeX document with the image:
image

2. Time to export an already-created image is negligible. With `rmsg` on, the time to export an SVG or PDF never reached 0.01s, for PNG it was between 0.02 and 0.04s. So the marginal time cost of exporting the PDF as well should be negligible.
3. Image sizes are small. For the export of `scatter price mpg`, the PDF was 9.7KB, the SVG was 16KB, and the PNG was 22KB.

kylebarron:
- [x] closes #175
- [ ] tests added / passed
- [ ] whatsnew entry

This returns the PDF along with the SVG by default, with the redundancy for PNGs off by default. It warns the user if one of the graph file sizes are more than 2MiB in size.

### DIFF
--- a/docs/src/using_stata_kernel/intro.md
+++ b/docs/src/using_stata_kernel/intro.md
@@ -12,6 +12,34 @@ To minimize false positives, the graph keyword must appear at the beginning of a
 
 To display graphs from user created commands (e.g., `coefplot`), use `graph display` after the command that generated the graph.
 
+### Graph display format
+
+`stata_kernel` must export the image in some format in order to load and display it. Sadly, there are considerable pros and cons to each image format:
+
+- `png` files can't be created with the console version of Stata, and thus are off limits to the Linux and Mac (console) modes of `stata_kernel`. Additionally, they can look pixelated when scaled up.
+- `svg` files can be created by Stata 14 and 15 on all platforms, and look crisp at all sizes, but can't be used with LaTeX PDF output.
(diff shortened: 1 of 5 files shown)


kylebarron:
I think this is good to merge but I want to merge #184 first.

PR #184, opened 2018-09-16
Simplify %set magic

### LINKED ISSUE #182: Let %set use the same keys as the configuration
@mcaceresb 
I think 
```
%set graph_format svg
```

is more straightforward to use than
```
%set graph --format svg
```

because `graph_format` is the key already used in the configuration file and listed in the [documentation](https://kylebarron.github.io/stata_kernel/getting_started/#graph-settings).

kylebarron:
- [ ] closes #182
- [ ] tests added / passed
- [ ] whatsnew entry

@mcaceresb Can you test this out, since this changes your code?

Instead of bifurcating `%set` into `%set graph ...` and `%set _all ...`, this uses just `%set key value [--permanently] [--reset]`. I think this is simpler for the user and simpler to manage.

This also changes the autocompletions for `%set` to use the list of setting names.

### DIFF
--- a/docs/src/getting_started.md
+++ b/docs/src/getting_started.md
@@ -4,7 +4,7 @@ It doesn't take much to get `stata_kernel` up and running. Here's how:
 
 ## Prerequisites
 
-- **Python**. In order to install the kernel, Python >= 3.5 needs to be installed on the computer on which Stata is running.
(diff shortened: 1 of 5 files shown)


mcaceresb:
Looks good.

PR #195, opened 2018-09-20
[WIP] File path autocompletion

### LINKED ISSUE #119: Autocomplete for file paths
IPython does this within strings. It might be straightforward to replicate the behavior here.

kylebarron:
- [x] closes #119 

Work in progress:

- [x] Store values of globals in memory
- [x] Determine when a string is a path
- [x] Substitute globals with their value
- [x] Use os.listdir
- [x] Check Windows support


kylebarron:
@mcaceresb 	what are specific Stata commands that take file paths? I'm thinking that might be preferable to always suggesting file paths in the global scope

kylebarron:
`save`, `saveold`, `use`, `graph export`, `graph use`, and always after `using`. Do you think that's better than suggesting file paths everywhere?

mcaceresb:
Technically any user program could take a file path as input---certainly any program that parses "using" would.

If you want to restrict it to at least be able to test it out, I'd say it should start with use, save, or using anywhere in the program call.

For brownie points add stuff like mkdir, erase, etc. I think that in theory this should work im the global scope, but it's better to have some support than not any at all.

mcaceresb:
> save, saveold, use, graph export, graph use, and always after using.

I think anywhere after save, use, or using text with. slash should trigger file path completion.

> Do you think that's better than suggesting file paths everywhere?

Not really, but filepaths everywhere is a marginal improvement in my mind. I think your idea is good and filepaths everywhere might be too complicated for what it actually buys.

kylebarron:
I've gotten some progress with the following caveats:

- Relative paths are calculated from Python's current working directory, not Stata's
- relative paths from the current directory are given suggestions that start with `./`
- Works with global macros but then substitutes the value of the global macro when returning the suggestion

kylebarron:
It's now also case insensitive. @mcaceresb 	want to try it out?

mcaceresb:
Pretty cool. No spaces yet, right? Eventually it might make sense to make a distinction between files and directories.

screenshot_20181021_105811

kylebarron:
?
I do... The directories have `/` at the end

kylebarron:
But yeah no spaces support. It'll suggest files with spaces as long as what you've typed so far has no spaces

mcaceresb:
> I do... The directories have / at the end

No, I mean that commands like `cd` can't take files as input.

kylebarron:
I mean it seems like there wouldn't be that many commands that can't take files as input

kylebarron:
With the last couple fixes, I think this should be stable, including Windows.

PR #217, opened 2018-10-25
Refactor `%browse` to use the same code as `%head`

### LINKED ISSUE #215: %browse sends previous dataset even when no data in memory
To reproduce:
```stata
sysuse auto
%browse
clear
%browse
```

Note that `%head` doesn't have the same issue.

kylebarron:
Supersedes https://github.com/kylebarron/stata_kernel/pull/89.

Fixes #215

### DIFF
--- a/stata_kernel/ado/_StataKernelHead.ado
+++ b/stata_kernel/ado/_StataKernelHead.ado
@@ -1,6 +1,6 @@
 capture program drop _StataKernelHead
 program _StataKernelHead
-    syntax [anything] [if] using, [*]
+    syntax [anything] [if] using, [N_default(int 10) *]
     set more off
     set trace off
     if ( !`=_N > 0' ) error 2000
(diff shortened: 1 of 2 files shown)


kylebarron:
This is way easier than having separate code for `%browse` and `%head` when they do the same thing.
@mcaceresb

stata_kernel/ado/_StataKernelHead.ado
```diff
@@ -15,7 +15,7 @@ program _StataKernelHead
         if ( _rc ) error 198
         local n = `n1'
     }
-    else local n = 10
+    else local n = `N_default'
```
mcaceresb: I don't think this works bc uppercase is a shortcut. Should be `n_default`

kylebarron: oh lol yeah

kylebarron:
Also, this is hitting the same issue as I mentioned in #89. Using `sysuse auto`, `%browse` says
```
error: observation numbers out of range
```
I think this is because I'm asking for N from 1/200, and there aren't 200 obs

mcaceresb:
`local n = min(_N, 200)` ?

kylebarron:
Seems to work good for me!

PR #228, opened 2018-11-05
Hide (most) Stata kernel output from user logs.

### LINKED ISSUE #194: use a different log file for _StataKernelCompletions
Might solve #154. Originally mentioned here: https://github.com/kylebarron/stata_kernel/issues/119#issuecomment-420367752

mcaceresb:
- [ ] closes #
- [ ] tests added / passed
- [ ] whatsnew entry

**EDIT:** On reflection, I don't think this closes #194 bc it doesn't redirect the output, but it _would_ produce a much cleaner user log.


kylebarron:
It's cleaner... I still see this:

```
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
      name:  <unnamed>
       log:  /home/kyle/test.log
  log type:  text
 opened on:   5 Nov 2018, 15:07:15

. `7a403a9adb587f9763c9fdb2f36e84a0'
. _StataKernelLog off

. `d450cfde56d6cad2042c8f0870cc76f5'
. sysuse auto, clear
(1978 Automobile Data)

. `0fb5dfbcd6a4c91ccf6670689cc566cb'
. _StataKernelLog off

. `d450cfde56d6cad2042c8f0870cc76f5'
. reg price mpg

      Source |       SS           df       MS      Number of obs   =        74
-------------+----------------------------------   F(1, 72)        =     20.26
       Model |   139449474         1   139449474   Prob > F        =    0.0000
    Residual |   495615923        72  6883554.48   R-squared       =    0.2196
-------------+----------------------------------   Adj R-squared   =    0.2087
       Total |   635065396        73  8699525.97   Root MSE        =    2623.7

------------------------------------------------------------------------------
       price |      Coef.   Std. Err.      t    P>|t|     [95% Conf. Interval]
-------------+----------------------------------------------------------------
         mpg |  -238.8943   53.07669    -4.50   0.000    -344.7008   -133.0879
       _cons |   11253.06   1170.813     9.61   0.000     8919.088    13587.03
------------------------------------------------------------------------------

. `f0eed606d0953e83e1bf4fec3a644b4f'
. _StataKernelLog off

. `d450cfde56d6cad2042c8f0870cc76f5'
. log close
      name:  <unnamed>
       log:  /home/kyle/test.log
  log type:  text
 closed on:   5 Nov 2018, 15:07:24
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

mcaceresb:
Right; change title to [WIP]?

I wasn't 100% how to hide the hashes because they get executed before I can intercept them; likewise once I turn the logs back on the hash for that command gets appended.  I suppose I could do that in python.

TODO: Something like

```python
res = re.sub("{0}($|\r?\n)".format(md5), '', res)
```

mcaceresb:
TODO: `mata` code break inserts both an md5 hash and a pair of `{}`. This does not get cleaned and must look odd form the user POV.

kylebarron:
Is this something that will have to be changed any time what's written to the log is changed?

mcaceresb:
TODO: The `include ~/.stata_kernel_cache/include.do` are still in the logs.

mcaceresb:
> Is this something that will have to be changed any time what's written to the log is changed?

What do you mean?

kylebarron:
Like if I add another command to `kernel.post_do_hook()` that writes output to Stata. Then would I later have to figure out a new regular expression or something for each text that might get written to Stata output?

mcaceresb:
Interesting. In `post_do_hook` as long as you put stuff between `cleanLog('off')` and `cleanLog('on')` the answer is no.

I also moved the md5 cleaning code to the end of `StataKernel.do` so it gets run after anything gets executed and you don't have to remember to add it if you add another `stata.do` statement somewhere.

The first `_StataKernelLog off` is shown because when a log file is open, the kernel doesn't know it is open so it doesn't know to clean it.

kylebarron:
Ok. I hadn't looked into the details of this PR yet.

PR #229, opened 2018-11-06
Completions on boundaries, e.g. gen a = (x<tab>

mcaceresb:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

Completions now work on word boundaries, not just spaces; e.g. `gen a = (x<tab>` will now suggest variables and paths starting with`x`.

### DIFF
--- a/stata_kernel/completions.py
+++ b/stata_kernel/completions.py
@@ -152,7 +152,9 @@ class CompletionsManager():
         # Detect space-delimited word.
         env = 0
         env_add = 0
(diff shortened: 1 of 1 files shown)


kylebarron:
Does this make it easier to show quoted paths with spaces?

mcaceresb:
No. I guess it could be retooled for that? I kept typing stuff of the form

```
foo(var<tab>
```

Nothing would happen and I thought it was annoying.

PR #246, opened 2018-11-28
Fixes #245: Send image as both SVG and HTML wrapped within iframe

### LINKED ISSUE #245: Displaying images no longer works in qtconsole
#### Problem description

Merge request #235 (in an attempt to fix #234) broke image display in `qtconsole`. I don't think there is a universal fix. Perhaps an option to wrap SVG in an iframe? The double display thing would not really a problem in qtconsole bc not everything is wrapped up in HTML.

#### Debugging log

Everything is working as designed, but qtconsole can't display HTML.

#### Code Sample

```stata
sysuse auto
hist price
```

#### Expected Output

A plot displayed.

#### Other information

- Linux
- Stata version 15.1
- Package version 1.7.4

mcaceresb:
- [X] closes #245
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/stata_kernel/kernel.py
+++ b/stata_kernel/kernel.py
@@ -221,6 +221,11 @@ class StataKernel(Kernel):
                     'width': width,
                     'height': height}
 
+                content['data']['image/svg+xml'] = img
+                content['metadata']['image/svg+xml'] = {
+                    'width': int(root.attrib['width'][:-2]),
+                    'height': int(root.attrib['height'][:-2])}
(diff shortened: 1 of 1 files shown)


kylebarron:
Awesome. I just checked and the notebook and lab environments both choose the HTML display over the SVG display, which is good. I was worried they might display SVG over HTML when both are sent, in which case this would negate the previous fix, but it looks good.

PR #266, opened 2018-12-17
Cleaning logs no longer clears r()

### LINKED ISSUE #265: r() class deleted
#### Problem description

Sample code:
```stata
sysuse auto
sum mpg
local res = r(max)
di `res'
```

In Stata console:
image

In Jupyter:
image

It actually works when you paste it all in at once:
image

I'm guessing the attempt to clean the log discards the `r()` state in Stata.

mcaceresb:
- [X] closes #265
- [x] tests added / passed
- [ ] whatsnew entry

Not sure if this is the most elegant way to deal w the issue (not crazy about how many "captures" I'm doing), but certainly on the easier side. Tested out your example and it worked fine. Let me know if/how I'm supposed to add tests for this.


kylebarron:
FYI good to run tests now when doing PRs. Just `pytest` from project directory.

mcaceresb:
This fails on the master branch on my computer. `S_StataSE` doesn't exist in IC (it does in SE), and `S_StataMP` doesn't exist in SE or IC (it does in MP, as does `S_StataSE`).

[pytest.log](https://github.com/kylebarron/stata_kernel/files/2688780/pytest.log)

kylebarron:
I updated the structure of the test files, and updated that test specifically. You should be able to run `pytest` and run all the tests now.

I've moved each part of the kernel tests into their own file. So if you just want to test completions, you can run `pytest tests/test_kernel_completions.py`. Each file subclasses the test framework in [`tests/utils.py`](https://github.com/kylebarron/stata_kernel/blob/86be7b27729514278a29198b863b5823f2dfea63/tests/utils.py), which in turn subclasses `jupyter_kernel_test`. I've added some helper functions to run code in `tests/utils.py`, and I think that will be further expanded with more helper functions in the future.

PR #294, opened 2019-02-10
Code introspection now finds the last word, adds mata support.

mcaceresb:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

Previously, mata functions were not parsed. Further, this introspects the last keyword word, as opposed to the first, which can be confusing.

### DIFF
--- a/stata_kernel/kernel.py
+++ b/stata_kernel/kernel.py
@@ -375,15 +375,33 @@ class StataKernel(Kernel):
 
     def do_inspect(self, code, cursor_pos, detail_level=0):
         inspect_keyword = re.compile(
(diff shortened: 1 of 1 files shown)


mcaceresb:
Two fixes on top of mata:

```stata
gen x = regexr(
```

Now triggers the completion for `regexr` instead of `gen`. Further,

```stata
gen x = regexr(substr(subinstr(
```

Triggers the completions for `regexr`, then `substr`, then `subinstr`.

kylebarron:
Completion or introspection?

kylebarron:
The introspection in qtconsole looks really good!

PR #324, opened 2020-01-05
Removed deprecation

Mioroxi:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry
 replace=True

Just as the title says, it's a tiny edit that resolved an issue which stressed me out for 20 mins trying to find a workaround for. Everything else works as intended

### DIFF
--- a/stata_kernel/install.py
+++ b/stata_kernel/install.py
@@ -37,7 +37,7 @@ def install_my_kernel_spec(user=True, prefix=None):
 
         print('Installing Jupyter kernel spec')
         KernelSpecManager().install_kernel_spec(
-            td, 'stata', user=user, replace=True, prefix=prefix)
+            td, 'stata', user=user, prefix=prefix)
 
 
 def install_conf(conf_file):


mcaceresb:
Sorry, what workaround is this for? That is, what is the issue that removing `replace=True` intending to fix?

kylebarron:
Looks like this:
https://github.com/jupyter/jupyter_client/blob/44980c13680f4e4226cf25f199ce4e4bb6e11296/jupyter_client/kernelspec.py#L326-L331

The user probably sees the deprecation notice upon installing

Mioroxi:
I got drunk as fuck and started doing shit, my bad.
That isn't what the workaround was for, since the program didn't install at all for me. I got a deprecation warning as well as a long list of python compile execute errors when I ran the installer, and then I narrowed down the problem to install.py. 
All I remember is that after removing replace=True I've had no issues whatsoever.
Sorry again for the awkwardness

kylebarron:
Given the deprecation warning, I'm happy to merge this @mcaceresb . Any thoughts?

PR #328, opened 2020-02-08
[DRAFT] Add support for a global configuration file.

Carreau:
Use '/etc/stata_kernel.conf' to pull default values from; this should
allow simpler deployment in JupyterHub environment that usually are on
linux.

This avoid to create a configuration file on a per-user basis.

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

-- 

Draft, I need to try it at work where stata is installed.

### DIFF
--- a/docs/src/using_stata_kernel/configuration.md
+++ b/docs/src/using_stata_kernel/configuration.md
@@ -20,6 +20,11 @@ If you want these changes to be stored permanently, add `--permanently`:
 %set graph_format png --permanently
 ```
 
+!!! info "System wide configuration file for JupyterHub"
+
+    If you are installign `stata_kernel` in Jupyter Hub you must create a system
+    wide configuration file in `/etc/stata_kernel.conf` to provide default
+    values.
 
 ## General settings
 
--- a/stata_kernel/config.py
+++ b/stata_kernel/config.py
@@ -7,6 +7,8 @@ from configparser import ConfigParser, NoSectionError
 
 from .utils import find_path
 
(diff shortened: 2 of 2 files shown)


mcaceresb:
I'll have to check the behavior of `ConfigParser().read()` if the file doesn't exist. My instinct is that this should have some conditional logic to only do this if `platform.system()` is not Windows or Darwin, though. You reckon it makes more sense to have it as is?

kylebarron:
As with @mcaceresb ; what happens if `/etc/stata_kernel.conf` doesn't exist?

```py
        for c in (global_config, self.config):
```
So user config values overwrite global config values, right?

Since the user config is still the one referenced by `self.config`, I think that any values set by `%set` by the user would get updated correctly in their personal config.

mcaceresb:
@kylebarron `ConfigParser().read()` says "files that cannot be opened are silently ignored" so it should be fine (and my understanding of this line in the doc is that it won't affect other systems). Since `global_config` is only used in `__init__` then nothing from it will permeate downstream if the user updates their own config file or uses `%set`

Do see value in trying to provide an OS specific default for `GLOBAL_PATH`? If not, I'll merge the PR (though I'll change the internal docstring to also note that GLOBAL_PATH is hard-coded in this case).

kylebarron:
> Do you see value in trying to provide an OS specific default for GLOBAL_PATH?

You mean a system wide config location for other OSes? I don't think that's necessary. If someone asks for it we can come back to this issue and add it easily

Issue #309, opened 2019-07-22
Stata 16 Support


StevenLi-DS:
Just saw this new update in Stata. Is there any plan to support Stata 16 buddy?

https://www.stata.com/new-in-stata/#

**Update:**

```
Versions:
stata-kernel: 1.10.5
jupyterlab: 1.1.4
notebook: 5.7.8
Stata: Stata IC 16
OS: Windows 10

Paths:
Stata: C:\Program Files\Stata16
.stata_kernel.conf: C:\Users\your_user_name
```
I was using Stata 15 and installed new Stata 16 after uninstalling version 15 and did the following:

1. register Stata 16 (i.e., link the Stata Automation library) via the step in the [doc](https://kylebarron.dev/stata_kernel/getting_started/#prerequisites):

2. change the Stata path in `.stata_kernel.conf`:
`stata_path = C:\Program Files\Stata16\StataIC-64.exe`

I tested with the [example notebook](https://nbviewer.jupyter.org/github/kylebarron/stata_kernel/blob/master/examples/Example.ipynb) and everything works fine with Stata 16

lordflaron:
Not a developer, but I've used the kernel quite a bit. Have you tried just changing the executable in the config to point to Stata16? 

I wouldn't think that the versions would differ so much that Stata 16 wouldn't work.

mcaceresb:
Well, as far as I cam tell the only issue would be python support, but I don't see that anything else in Stata 16 would be a problem for the kernel.

StevenLi-DS:
@lordflaron haven't got my copy of it and will give it a shot once I get it.\

@mcaceresb Have you tried with 16? Is there anything I should be aware of?

mcaceresb:
@StevenLi-DS I don't have Stata 16. I just meant that going through their "new in stata" page the only thing that would trip up the kernel is python support. (Maybe some new keywords won't be highlighted, but that's second order.) I do want to add python support to the kernel, in the style of mata support, but since I don't have access to Stata 16 that will have to wait for now.

SiSeUniBE:
I'm using Stata 16 from within Atom via the stata_kernel already for some days now, and it works fine. As @mcaceresb already said, the execption is the integration of python into stata, which is not supported by the stata_kernel (it get stuck, if you try to use it). Great to hear that @mcaceresb plans to add python support to the kernel!
[...]

PR #330, opened 2020-02-10
Add Stata 16 to Windows default search list

kylebarron:
Ref: https://github.com/kylebarron/stata_kernel/issues/309#issuecomment-584343298

@mcaceresb I don't think this necessarily needs to be tested; do you ever use Stata on Windows?

### DIFF
--- a/stata_kernel/utils.py
+++ b/stata_kernel/utils.py
@@ -47,8 +47,9 @@ def win_find_path():
     import winreg
     reg = winreg.ConnectRegistry(None, winreg.HKEY_CLASSES_ROOT)
     subkeys = [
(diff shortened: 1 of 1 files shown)


mcaceresb:
I very occassionally fire up Stata 13 on a Windows VM, so I can't really test this properly. But I agree it doesn't need to be tested.

kylebarron:
And you wouldn't be able to test this anyways unless it was a Windows VM with Stata 16. Let me know if you have any other quick additions you want to make; otherwise I'll push this to PyPI as a patch.

mcaceresb:
No obj; cool.

PR #346, opened 2020-04-21
Support EPS graphics via user-specified program

### LINKED ISSUE #345: PS and EPS graphics formats
I still love the kernal, makes my PhD right now!!

I just recognized that the current version locks me out regarding graphics. I am running Stata 13 under Linux. This leaves me without any choice for graphics as Stata 13 manual reads:

> ps and eps are available for all versions of Stata; png and tif are available for all versions of Stata except Stata(console) for Unix; pdf is available only for Stata for Windows and Stata for Mac; and wmf and emf are available only for Stata for Windows.

Thus, I can only use ps and eps with my configuration, which is not supported by the kernel right now. Can you make this happen?

mcaceresb:
- [X] closes #345
- [ ] tests added / passed
- [ ] whatsnew entry

In #345 support for EPS was requested. This allows the user to specify EPS graphics if they can convert EPS to a PNG via a command line program. For example, on Linux the user can specify in `~/.stata_kernel.conf`

```
graph_format = eps
graph_epstopng_program = convert -density 300 {0} -resize '900x600' {1}
```

To use EPS from Stata and convert the image to `png` (in this case using `imagemagick`'s `convert` program).

### DIFF
--- a/stata_kernel/code_manager.py
+++ b/stata_kernel/code_manager.py
@@ -269,6 +269,8 @@ class CodeManager():
             pdf_dup = config.get('graph_svg_redundancy', 'True')
         elif graph_fmt == 'png':
             pdf_dup = config.get('graph_png_redundancy', 'False')
+        elif graph_fmt == 'eps':
+            pdf_dup = config.get('graph_eps_redundancy', 'False')
         pdf_dup = pdf_dup.lower() == 'true'
 
         dim_str = " width({})".format(int(graph_width * graph_scale))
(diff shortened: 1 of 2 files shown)


kylebarron:
@mcaceresb I forgot about this. You're free to merge if you want.

kylebarron:
Awesome. I should try to publish a new patch release to pypi

kylebarron:
Might be good to add something to the docs about `graph_epstopng_program`

PR #353, opened 2020-07-22
Add environmental variable option for config global and user locations

### LINKED ISSUE #347: Solution to "stata_path option in configuration file is missing or invalid"
Not sure if this belongs here, as this is my first GitHub post.

Anyways, I was having a problem running the "python -m stata_kernel.install" portion of the installation and my solution was not really covered in the FAQ.  Maybe it is, but it wasn't clear.

These are the EXTRA steps you need to take (i.e., you NEED to do the make shortcut, /register, etc.; but if you do that and get "stata_path option..." then try this):

1) Navigate to your personal folder: typically C:\Users\USERNAME (username is your username, so if your name is "statamcstataface" the directory would be C:\Users\statamcstataface)

2. Download the attached file (.stata_kernel.conf.txt) to your personal folder. 
[.stata_kernel.conf.txt](https://github.com/kylebarron/stata_kernel/files/4593625/default.stata_kernel.conf.txt)

3. Open the file.  On line two, rename the stata_path section - where it says [Insert the correct directory here] to the correct address of your STATA installation folder (make sure to delete the [ and ] as well).  I'm running StataMP-64, change that the StataSE if that's what your system is using.

4. Save as, and remove the ".txt" ending.  

The final name of the file should be .stata_kernel.conf and line two of said file should be your Stata installation folder. Then run "python -m stata_kernel.install" again. Hope it works for you, took me hours to figure out the problem.

If you don't want to download random files, here's the contents. Just make a new .txt file in your personal folder and paste it in:

[stata_kernel]
stata_path = "C:\Program Files\[Insert the correct directory here]\StataMP-64.exe"
execution_mode = automation
cache_directory = ~/.stata_kernel_cache
autocomplete_closing_symbol = False
graph_format = svg
graph_scale = 1
user_graph_keywords = coefplot,vioplot

gaulinmp:
- [x] Add environmental variable option for specifying config file location:
   - [x] Global config path can come from `STATA_KERNEL_GLOBAL_CONFIG_PATH`
   - [x] User config path can come from `STATA_KERNEL_USER_CONFIG_PATH`
- [x] Update documentation to reflect these changes
- [x] Add example config to end of documentation Configuration page.
   - [x] closes #347 

Example:

``` bash
$ STATA_KERNEL_USER_CONFIG_PATH=/opt/python/share/source/stata_kernel/.stata_kernel.conf python -c "from stata_kernel.config import config;print('config_path:', config.config_path, '\nstata_path:', config.get('stata_path'))"
config_path: /opt/python/share/source/stata_kernel/.stata_kernel.conf
stata_path: /opt/stata/stata

### DIFF
--- a/docs/src/using_stata_kernel/configuration.md
+++ b/docs/src/using_stata_kernel/configuration.md
@@ -24,7 +24,7 @@ If you want these changes to be stored permanently, add `--permanently`:
 
 !!! info "System wide configuration file for JupyterHub"
 
-    If you are installign `stata_kernel` in Jupyter Hub you must create a system
+    If you are installing `stata_kernel` in Jupyter Hub you must create a system
     wide configuration file to provide default values. The default location
     is in `/etc/stata_kernel.conf`, or defined by the environmental variable
     `STATA_KERNEL_GLOBAL_CONFIG_PATH`.
--- a/docs/src/using_stata_kernel/configuration.md
+++ b/docs/src/using_stata_kernel/configuration.md
@@ -1,7 +1,9 @@
 # Configuration
 
 The configuration file is a plain text file named `.stata_kernel.conf` and is
-located in your home directory. You can change any of the package's settings by
+located in your home directory, or defined by the environmental variable
+`STATA_KERNEL_USER_CONFIG_PATH`. Settings must be under the heading
+`[stata_kernel]`. You can change any of the package's settings by
 opening the file and changing the `value` of any line of the form
 
 ```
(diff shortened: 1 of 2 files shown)


gaulinmp:
I'm realizing that this won't work for the `python -m stata_kernel.install`, where the stata path would have to be defined in `~/.stata_kernel.conf`. Once it's installed any config file location specified through the environmental variables will work, but not on install.

I could add the same code to `install.py` to search environmental variables too, if that's useful.

kylebarron:
Any documentation updates are helpful, thanks for that.

Why is it important to allow other locations for the config file? Is it not enough to support the default locations in the global and user locations? I'm worried that this will make it more complicated for new users, who might not even understand what an environment variable is.

gaulinmp:
My use case was mostly not having access to /etc/ softlinking, so I put the global config file in the stata_kernel git repo folder, then added `export STATA_KERNEL_GLOBAL_CONFIG_PATH=/opt/python/share/source/stata_kernel/.stata_kernel.conf` to my .bashrc. But this wouldn't work for the user config path, because the kernel writes to that (if I'm understanding the code correctly). So I agree that the env config path for the user config might be more confusing than it's worth, since probably everyone has access to their own home directory.

I do think that having an example config file in the documentation is nice, so I could just push that change if the env var part isn't useful.

PR #357, opened 2020-09-22
Document #346 (support for EPS via user-specified program)

mcaceresb:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

As the title says, following [your comment](https://github.com/kylebarron/stata_kernel/pull/346#issuecomment-696497763). Let me know if the wording seems off; the only thing that changed is that `eps` was added to the docs as a format and `graph_epstopng_program` was added to the list of options..


kylebarron (approved):

kylebarron:
Thanks! New docs are published!

PR #364, opened 2020-10-12
Removed jupyter from requirements, modified docs

### LINKED ISSUE #361: Could jupyter be removed from the dependencies?
When working with conda environments it is usual to have jupyter (and jupyter lab, notebook, etc.) installed in one environment only (e.g. base), and only install ipykernel into the other environments that are to be used in jupyter notebook or lab. However for stata_kernel jupyter is listed as a requirement, therefore all the jupyter machinery will be installed when installing it into an environment.

Is jupyter really necessary, or would ipykernel be enough as a requirement? Removing it would allow for more lightweight conda envs.

stanmart:
- [x] closes #361 
- [ ] tests added / passed
- [ ] whatsnew entry

Added instructions to the docs for installing jupyter notebook/lab.

The conda-forge package should be up soon (waiting for the review from the conda-forge team), so also added instructions to install stata_kernel via conda.

The same tests fail as before the change:
test_kernel_display_data.py::StataKernelTestFramework::test_kernel_info - AttributeError: 'NoneType' obj...
test_kernel_stdout.py::TestKernelStdout::test_stata_more - AssertionError: 0 not greater than or equal to 1
test_kernel_stdout.py::StataKernelTestFramework::test_kernel_info - AttributeError: 'NoneType' object ha...

### DIFF
--- a/docs/src/getting_started.md
+++ b/docs/src/getting_started.md
@@ -33,7 +33,14 @@ conda install -c conda-forge stata_kernel
 python -m stata_kernel.install
 ```
 
-Otherwise, from a terminal or command prompt run:
+If Python 2 is the default ersion of Python on your system, you may need to use
+
+```bash
+conda install -c conda-forge stata_kernel
+python3 -m stata_kernel.install
+```
+
+If you do not use Anaconda/Miniconda, from a terminal or command prompt run:
 
 ```bash
 pip install stata_kernel
--- a/docs/src/getting_started.md
+++ b/docs/src/getting_started.md
@@ -26,7 +26,14 @@ It doesn't take much to get `stata_kernel` up and running. Here's how:
 
 ## Package Install
 
-To install the kernel, from a terminal or command prompt run:
+If you use Anaconda or Miniconda, from the Anaconda Prompt run:
+
+```bash
+conda install -c conda-forge stata_kernel
+python -m stata_kernel.install
+```
+
+Otherwise, from a terminal or command prompt run:
 
 ```bash
 pip install stata_kernel
@@ -37,6 +44,24 @@ If Python 2 is the default version of Python on your system, you may need to use
 ```bash
 pip3 install stata_kernel
 python3 -m stata_kernel.install
+```
+
+### Jupyter
+
+If you chose to install Anaconda you already have Jupyter and Jupyter Lab installed.
+
+If you downloaded Miniconda, you need to install Jupyter and optionally Jupyter Lab:
+
+```bash
+conda install jupyter
+conda install jupyterlab
+```
+
+Otherwise you can install them using `pip`:
+
+```bash
+pip install jupyter
+pip install jupyterlab
 ```
 
 In order to get syntax highlighting in Jupyter Lab, run:
@@ -51,9 +76,15 @@ If you didn't install Python from Anaconda, the `conda` command won't work and y
 
 To upgrade from a previous version of `stata_kernel`, from a terminal or command prompt run
 
+```bash
+conda update stata_kernel -y
 ```
+in the case of Anaconda/Miniconda or
+
+```bash
 pip install stata_kernel --upgrade
 ```
+otherwise.
 
 When upgrading, you don't have to run `python -m stata_kernel.install` again.
 
--- a/docs/src/getting_started.md
+++ b/docs/src/getting_started.md
@@ -48,21 +48,35 @@ python3 -m stata_kernel.install
 
 ### Jupyter
 
-If you chose to install Anaconda you already have Jupyter and Jupyter Lab installed.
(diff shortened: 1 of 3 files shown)


docs/src/getting_started.md
```diff
@@ -26,7 +26,14 @@ It doesn't take much to get `stata_kernel` up and running. Here's how:
 
 ## Package Install
 
-To install the kernel, from a terminal or command prompt run:
+If you use Anaconda or Miniconda, from the Anaconda Prompt run:
+
+```bash
+conda install -c conda-forge stata_kernel
+python -m stata_kernel.install
```
kylebarron, line 33: probably good to suggest `python3` as below

docs/src/getting_started.md
```diff
@@ -39,6 +46,24 @@ pip3 install stata_kernel
 python3 -m stata_kernel.install
 ```
 
+### Jupyter
+
+If you chose to install Anaconda you already have Jupyter and Jupyter Lab installed.
+
+If you downloaded Miniconda, you need to install Jupyter and optionally Jupyter Lab:
```
kylebarron, line 53: I'd say Jupyter **or** Jupyter Lab, there's no reason both are needed. And I'd suggest a slight nudge towards Jupyter lab

kylebarron: Also you can probably suggest installing `notebook`, not `jupyter`.

docs/src/getting_started.md
stanmart: AFAIK for Anaconda/Miniconda distributions the name of the executable is always `python`, regardless of OS or other versions present. When someone runs it from the anaconda prompt it should always find the correct python executable. Added pip3 to the later instructions for installing notebook/jupyterlab, though.

docs/src/getting_started.md
stanmart: Thanks, agreed.

docs/src/getting_started.md
```diff
@@ -48,29 +48,43 @@ python3 -m stata_kernel.install
...
 
 ```bash
-pip install jupyter
-pip install jupyterlab
+pip install jupyterlab 
+pip3 install jupyterlab  # if Python 2 is the default version
 ```
 
+If you would not like to install Jupyter Lab and only need the Notebook, you can install it by running
+
```
stanmart, line 67: Actually, this section might not be necessary at all. As this section of the docs is aimed at people new to Jupyter, it might make sense to just provide instructions for installing the lab, which will also install the notebook in turn as it is a dependency of the lab. The text at the beginning still makes it clear that the Notebook is enough in itself.

Having installation instructions or the notebook as well makes the matter a bit clearer, but maybe longer than it needs to be. What do you think?

stanmart:
I also got this feedback from a conda-forge team member:

> While twine is listed in setup_requires this is a really weird requirement and doesn't actually seem to be used for setup (or in the entire code base). I think it's just listed as a convenience for those building packages. Can you remove?

Should it possibly be removed from `setup_requires` and added to `requirements_dev.txt` instead?

docs/src/getting_started.md
kylebarron: With Anaconda, `python` and `python3` are both symlinks to the latest release:
```
> ll ~/local/anaconda3/bin/python*
lrwxr-xr-x  1 kyle  staff     9B Jun  8 16:08 /Users/kyle/local/anaconda3/bin/python -> python3.7
lrwxr-xr-x  1 kyle  staff     9B Jun  8 16:08 /Users/kyle/local/anaconda3/bin/python3 -> python3.7
-rwxrwxr-x  1 kyle  staff   4.0M Jun  8 16:07 /Users/kyle/local/anaconda3/bin/python3.7
```

I agree that `python3` isn't usually necessary, but I'm trying to think about the user who's new to Python, and especially mac users whose default `python` points to the `/usr/bin/python` 2.7 might get confused.

stanmart: Ah, I did not now this (I use Windows and there is no python3 symlink there). Will do the change.

stanmart:
The conda recipe is up and available on conda-forge: [https://github.com/conda-forge/stata_kernel-feedstock](https://github.com/conda-forge/stata_kernel-feedstock)
I tested it and it seems to work fine (as it should, it is a pure Python package after all). Please let me know if I should add you as a recipe maintainer or if you see any issues with it.

docs/src/getting_started.md
```diff
@@ -39,6 +46,24 @@ pip3 install stata_kernel
...
+conda install jupyter
+conda install jupyterlab
+```
+
+Otherwise you can install them using `pip`:
+
+```bash
+pip install jupyter
+pip install jupyterlab
+```
```
kylebarron: I don't think you need to suggest installing `jupyter`, since that's just the metapackage. `jupyterlab` should install everything necessary.
`````suggestion
```bash
conda install jupyterlab
```

Otherwise you can install them using `pip`:

```bash
pip install jupyterlab
```
`````

kylebarron: Oh sorry I accidentally wasn't looking at the most recent commit...

PR #377, opened 2021-01-05
Added stack trace to kernel import

mcaceresb:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

The title. I think the full traceback when debugging might be better than the error message.

Flake complains about bare "except" and importing the kernel but not using it, btw.

### DIFF
--- a/stata_kernel/__init__.py
+++ b/stata_kernel/__init__.py
@@ -1,8 +1,11 @@
 """An example Jupyter kernel"""
 
+import traceback
+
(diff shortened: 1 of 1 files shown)


kylebarron (approved):

kylebarron:
I think you need to merge master

mcaceresb:
@kylebarron Yup, sorry. Did that.

PR #363, opened 2020-10-12
Fix install on linux when can't find stata path

kylebarron:
Closes #362

Traceback when trying to install on an ubuntu machine without Stata:

```
> python -m stata_kernel.install
Traceback (most recent call last):
  File "/home/ubuntu/local/miniconda3/lib/python3.8/runpy.py", line 185, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/home/ubuntu/local/miniconda3/lib/python3.8/runpy.py", line 111, in _get_module_details
    __import__(pkg_name)
  File "/home/ubuntu/local/miniconda3/lib/python3.8/site-packages/stata_kernel/__init__.py", line 5, in <module>
    from .kernel import StataKernel
  File "/home/ubuntu/local/miniconda3/lib/python3.8/site-packages/stata_kernel/kernel.py", line 16, in <module>
    from .config import config
  File "/home/ubuntu/local/miniconda3/lib/python3.8/site-packages/stata_kernel/config.py", line 171, in <module>
    config = Config()
  File "/home/ubuntu/local/miniconda3/lib/python3.8/site-packages/stata_kernel/config.py", line 86, in __init__
    self.raise_config_error('stata_path')
  File "/home/ubuntu/local/miniconda3/lib/python3.8/site-packages/stata_kernel/config.py", line 161, in raise_config_error
    raise ValueError(dedent(msg))
ValueError: stata_path option in configuration file is missing or invalid
Refer to the documentation to see how to set it manually:

https://kylebarron.dev/stata_kernel/using_stata_kernel/configuration/
```

and `~/.stata_kernel.conf` isn't created. It looks like importing the kernel fails because of a bug in the config.

PR #379, opened 2021-02-25
Simplify kernel installation for conda packaging

mcg1969:
- [ ] closes #xxxx
- [ ] tests added / passed
- [X] whatsnew entry

The conda package for `stata_kernel` offered by conda-forge requires a manual step following installation. This is, in theory, not necessary. It should be possible to do something like `python -m stata_kernel.install --sys-prefix` within the conda recipe itself, so that the kernel specification is included within the tarball. Other kernels successfully do this.

I believe this PR will accomplish this for the Stata kernel. To accomplish this, we've done a couple of things:
- Remove the `from . import kernel` from `__init__.py`. This allows the conda recipe to complete the packaging without importing the entirety of the `stata_kernel` dependency set. Of course, this also means that the `kernel.json` file has to call `stata_kernel.kernel` directly instead of just `stata_kernel`.
- Modify the installation process to reduce its import requirements as well. To accomplish this, I added a new installation flag `--no-conf-file` that skips the creation of the conf file, and thereby bypassing even the need to look for a Stata binary.

### DIFF
--- a/stata_kernel/__init__.py
+++ b/stata_kernel/__init__.py
@@ -3,9 +3,3 @@
 import traceback
 
 __version__ = '1.12.2'
-
-try:
-    from .kernel import StataKernel
-except:
-    print('Cannot import kernel')
-    traceback.print_exc()
--- a/stata_kernel/__init__.py
+++ b/stata_kernel/__init__.py
@@ -1,5 +1,3 @@
 """An example Jupyter kernel"""
 
-import traceback
-
 __version__ = '1.12.2'
--- a/stata_kernel/__init__.py
+++ b/stata_kernel/__init__.py
@@ -1,3 +1,3 @@
 """An example Jupyter kernel"""
 
-__version__ = '1.12.2'
+__version__ = '1.12.3'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [XXX] - 2021-02-25
+## [1.12.3] - 2021-02-25
 
 - Simplify kernel installation for conda environments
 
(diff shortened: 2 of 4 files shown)


kylebarron:
I don't see any big issues but I'd really defer to @mcaceresb because I haven't used this library for a couple years now.

stata_kernel/__init__.py
```diff
@@ -1,11 +1,3 @@
...
 
-import traceback
-
 __version__ = '1.12.2'
-
-try:
-    from .kernel import StataKernel
-except:
-    print('Cannot import kernel')
-    traceback.print_exc()
```
kylebarron, line 11: I'm curious if there will be a regression because of these changes. We've had some changes here lately https://github.com/kylebarron/stata_kernel/pull/363, https://github.com/kylebarron/stata_kernel/pull/377

mcaceresb (approved):
It looks OK to me, but I didn't have occasion to test it outside my own (Linux) laptop.

stata_kernel/__init__.py
mcaceresb: The StateKernel import is moved to `stata_kernel/__main__.py` instead of `stata_kernel/__init__.py`, but unsure whether the try/except block is still needed then. It seems harmless to have it so I added it back.

kylebarron:
Feel free to merge

stata_kernel/__init__.py
mcg1969: Ouch! I am sorry that for some reason this conversation did not make it to my inbox. The problem with having this here is that you now import `stata_kernel.kernel` even when you do not need to, which then imports dependencies that may not be present. In particular, you shouldn't necessarily need to important `stata_kernel.kernel` to do `stata_kernel.install`.

PR #394, opened 2021-05-16
add support for Stata 17 graphs

### LINKED ISSUE #391: graph display bug
#### Problem description

- This should explain **why** the current behavior is a problem and why the expected output is a better solution.
- **Note**: Many problems can be resolved by simply upgrading `stata_kernel` to the latest version. Before submitting, please try:

    ```
    pip install stata_kernel --upgrade
    ```
    and check if your issue is fixed.

#### Debugging log

If possible, attach the text file located at

```
$HOME/.stata_kernel_cache/console_debug.log
```
where `$HOME` is your home directory. This will help us debug your problem quicker.

**NOTE: This file includes a history of your session. If you work with restricted data, do not include this file.**

#### Code Sample

Especially if you cannot attach the debugging log, please include a [minimal, complete, and verifiable example.](https://stackoverflow.com/help/mcve)

```stata
use "https://stats.idre.ucla.edu/stat/stata/notes/hsb2", clear
scatter read math, title("Reading score vs Math score")
scatter math science, title("Math score vs Science score")

```

#### Expected Output
image

#### unexpected output
Just display like this, 

image

If you didn't attach the debugging log, please provide:

- Operating System win10
- Stata version 17.0
- Package version 1.12.2 (You can find this by running `pip show stata_kernel` in your terminal.)

simon-d-s:
- [x] closes #391
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/stata_kernel/stata_session.py
+++ b/stata_kernel/stata_session.py
@@ -299,7 +299,7 @@ class StataSession():
         error_re = r'^r\((\d+)\);'
 
         # The minimum linesize in Stata is 40 characters
-        g_exp = r'\(file {}'.format(self.cache_dir_str[:34])
+        g_exp = r'\(?file {}'.format(self.cache_dir_str[:34])
 
         more = r'^--more--'
         eol = r'\r?\n'
@@ -325,6 +325,9 @@ class StataSession():
                 continue
             if match_index == 2:
                 g_path = [self.expect_graph(child, child.match.group(0))]
+                if g_path[0] is None:
+                    res = None
+                    continue
                 if ((config.get('graph_format') == 'svg')
                         and config.get('graph_svg_redundancy', 'True')) or (
                             (config.get('graph_format') == 'png')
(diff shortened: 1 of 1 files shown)


stata_kernel/stata_session.py
```diff
@@ -299,7 +299,7 @@ def expect(self, text, child, md5, text_to_exclude=None, display=True):
         error_re = r'^r\((\d+)\);'
 
         # The minimum linesize in Stata is 40 characters
-        g_exp = r'\(file {}'.format(self.cache_dir_str[:34])
+        g_exp = r'\(?file {}'.format(self.cache_dir_str[:34])
```
kylebarron, line 302: Probably good to add a comment above this noting that the `(` doesn't exist in Stata 17

kylebarron (approved):

stata_kernel/stata_session.py
simon-d-s: Thanks for the feedback @kylebarron , I've added a comment in [97709b6](https://github.com/kylebarron/stata_kernel/pull/394/commits/97709b6c46c1a01c670283ebd05920270044187d)

stata_kernel/stata_session.py
```diff
@@ -325,6 +329,9 @@ def expect(self, text, child, md5, text_to_exclude=None, display=True):
                 continue
             if match_index == 2:
                 g_path = [self.expect_graph(child, child.match.group(0))]
+                if g_path[0] is None:
```
mcaceresb, line 332: @simon-d-s Sorry I've been MIA. Can this be fixed by instead adding `^` at the start of `g_exp`? Instead of the logic here and the `if else` logic below, add `^` at the start of `g_exp` so that the messages for a file not existing and the file being saved are not confused?

simon-d-s: @mcaceresb No problem.
For Stata 17 it would work, but I cannot see how it would work for Stata 16.

mcaceresb: @simon-d-s I tested it out on [this branch](https://github.com/kylebarron/stata_kernel/pull/392/files) and it was working for me on 15 and 16. I don't have 17 though.

simon-d-s: @mcaceresb Sorry, I did not fully understand what you mean - with the code it's clear now.
Unfortunately, this will not work on Stata 17. As you can see in the following example, the beginning the beginnings of the messages for a file not existing and the file being saved are not confused look differ only by the `(`: 
```
. sysuse auto
(1978 automobile data)

. hist price
(bin=8, start=3291, width=1576.875)

. graph export foobar.png, replace
(file foobar.png not found)
file foobar.png saved as PNG format
```

mcaceresb: @simon-d-s Ah, I see what you mean. I'll try to review your proposed fix when I get a sec. These weeks are a bit full for me, though.

PR #406, opened 2021-11-22
Use exactly one kernel_cache_dir and config object per kernel instance

### LINKED ISSUE #405: Sessions assume exclusive access to files in cache_directory, breaking simultaneous sessions
#### Problem description
Issue arises when a user runs two or more stata_kernel sessions simultaneously with the same `cache_directory` path configured. The stata_kernel uses the same `include.do` file in the `cache_dir` directory for both sessions. This leads to a race condition between sessions in writing to and reading/executing from `include.do`.

The effect is intermittent, seemingly inexplicable Stata errors resulting from Stata executing code run from a different session.

#### Code Sample
To reproduce, run multiple simultaneous sessions. Easiest to reproduce when the notebooks for each session have multiple cells that may start near-simultaneously.

#### Expected Output
No Stata errors.

#### Other information
- OS: RHEL 7, RHEL 8
- Stata: 16.0mp8c
- stata-kernel: 1.12.2

ryan-wallace-frbb:
- [x] closes #405
- [x] tests added / passed
- [x] whatsnew entry

whatsnew entry:
* fix intermittent Stata errors when running simultaneous sessions (#405)

### DIFF
--- a/stata_kernel/config.py
+++ b/stata_kernel/config.py
@@ -109,6 +109,10 @@ class Config():
         return self.env.get(key, backup)
 
     def set(self, key, val, permanent=False):
+        if key.startswith('cache_dir'):
+            val = Path(val).expanduser()
+            val.mkdir(parents=True, exist_ok=True)
+
         self.env[key] = val
 
         if permanent:
--- a/stata_kernel/config.py
+++ b/stata_kernel/config.py
@@ -4,6 +4,7 @@ import platform
 
 from pathlib import Path
 from textwrap import dedent
+from tempfile import TemporaryDirectory
 from configparser import ConfigParser, NoSectionError
 
 from .utils import find_path
@@ -77,9 +78,11 @@ class Config():
             except NoSectionError:
                 pass
 
-        cache_dir = Path(self.get('cache_directory',
-                                  '~/.stata_kernel_cache')).expanduser()
-        cache_dir.mkdir(parents=True, exist_ok=True)
+        cache_par_dir = Path(self.get('cache_directory',
+                                      '~/.stata_kernel_cache')).expanduser()
+        cache_par_dir.mkdir(parents=True, exist_ok=True)
+        self._cache_temp_dir = TemporaryDirectory(dir=str(cache_par_dir))
+        cache_dir = Path(self._cache_temp_dir.name)
 
         stata_path = self.get('stata_path', find_path())
         if not stata_path:
@@ -106,10 +109,6 @@ class Config():
         return self.env.get(key, backup)
 
     def set(self, key, val, permanent=False):
-        if key.startswith('cache_dir'):
-            val = Path(val).expanduser()
-            val.mkdir(parents=True, exist_ok=True)
-
         self.env[key] = val
 
         if permanent:
--- a/requirements_dev.txt
+++ b/requirements_dev.txt
@@ -1,6 +1,7 @@
 bumpversion==0.5.3
 coverage==4.5.1
 flake8==3.5.0
+importlib-metadata==0.23
 jupyter_kernel_test==0.3.0
 mkdocs-material==3.0.3
 mkdocs==1.0


mcaceresb:
@ryan-wallace-frbb It will take me a little bit to get to reviewing this. One conceptual question that I have: It seems to me that the solution, conceptually, is to instead of having the files saved directly into `cache_dir`, to make a sub-directory that is unique to a given kernel and save everything there.

This sounded relatively simple to me. I thought it would be sufficient was to add something like this:

```python
cache_dir = cache_dir / temp_dir_name
```

before `config.py` L82 (and perhaps make sure that the temp dir is not saved if `cache_directory` is saved interactively). However, your proposed changes look to be a lot more substantial than that. Can you please elaborate as to why the fix is more complex than my understanding of the problem would suggest? Thanks!

ryan-wallace-frbb:
Thanks, @mcaceresb.

The solution as implemented accords with your conceptual understanding--in config.py:84 we create a per-kernel temporary subdirectory of `cache_dir` to save all the runtime files for that kernel. 

But to keep the reference to this per-kernel `cache_dir` consistent within the kernel, we instantiate the `Config` class exactly once per kernel in kernel.py:69. Otherwise, each separate invocation of `config = Config()` in kernel.py:171 would yield a different randomly named `TemporaryDirectory` for the `cache_dir` every time the import `from .config import config` is run.

Nearly all the other modifications are to change references from `config` objects to the kernel attribute `self.conf`.

ryan-wallace-frbb:
Hi @mcaceresb, just a heads up that I addressed the issues that broke the previous builds. In particular, I made the `CodeManager` constructor take a `Config` object instead of a `Kernel` object to remove the dependency on Stata itself for the tests.

The latest build #825 now passes with Python 3.6 and 3.8. It fails on 3.5 and 3.7, but due to version issues with pip (as does the previous build #821).

ryan-wallace-frbb:
Hey @mcaceresb, I hope you're enjoying the holiday season.

Just a heads up that I made two slight modifications get this to pass the checks for all the Python versions tested. Please let me know if there's anything else I can do on this.

mcaceresb:
@ryan-wallace-frbb Thanks! Sorry I haven't had the time to look at this PR. I do hope to get to it soon (my semester technically is just finishing this week). Hope you are enjoying your break as well. Cheers.

mcaceresb:
@ryan-wallace-frbb Hey; I'm finally having some time to review this PR today/this week. Regarding [my previous comment](https://github.com/kylebarron/stata_kernel/pull/406#issuecomment-975936818), you had said that you said this:

> But to keep the reference to this per-kernel `cache_dir` consistent within the kernel, we instantiate the `Config` class exactly once per kernel in kernel.py:69. Otherwise, each separate invocation of `config = Config()` in kernel.py:171 would yield a different randomly named `TemporaryDirectory` for the `cache_dir` every time the import `from .config import config` is run.

Can you give me an example where `from .config import config` is run more than once? Your change seems to work (I still need to test it out a little more). However, if the only changes I keep are lines 7 and 84 in `config.py`, that seems to work too.

ryan-wallace-frbb:
Thanks @mcaceresb.

Previously, `config.py` ended with an instantiation of the `Config` class at the top-level of the module with `config = Config()`. I forgot that Python's import system ensures that each module is only imported once per interpreter session. So I mistakenly thought that a new instance of the `Config` class (with a corresponding new randomly named `cache_dir`) would be created each time `from .config import config` is run at the top of other modules in the package.

I agree now that it's sufficient to make the smaller set of changes you mentioned in `config.py`. I've updated this PR accordingly.

I think the only possibly unintuitive change now is creating an instance variable `_cache_temp_dir` for the `TemporaryDirectory` object in the `Config` class. This is done to maintain a reference to the `TemporaryDirectory` object itself so that it's not cleaned up automatically after the reference drops out of scope at the end of the class's `__init__` method. Since we use only the pathlib path to refer to the directory outside of the initialization, I prefixed the variable name with an underscore to indicate it's purpose is private to the class.

stata_kernel/config.py
```diff
@@ -106,10 +109,6 @@ def get(self, key, backup=None):
         return self.env.get(key, backup)
 
     def set(self, key, val, permanent=False):
```
mcaceresb, line 111: Why is this deleted? Wouldn't it be necessary for setting  the `cache_dir` interactively?

mcaceresb (commented):
This is what I initially thought would need to happen. Sorry it took so long to review!

Just one Q before I merge.

stata_kernel/config.py
ryan-wallace-frbb: Yes, I agree. I overlooked support for setting `cache_dir` interactively.

I've consolidated the logic for setting `cache_dir` into the `set` function so that it isn't duplicated in the `__init__` method.

mcaceresb: There are three issues with this:

- My kernel crashes when I try to set the directory and then run any command. Unsure why.
- Setting the cache directory interactively would now set it to have a temporary directory. If it's set permanently then the temporary directory would persist across sessions.
- I also don't think it's necessary to set a new temporary directory every time it's set interactively in the first place (though I can more or less see why it could be useful).

_However_, I think that your previous change was fine. If you revert and add these lines back in as they were I can merge it. Of course, feel free to try to fix the above issues, but as I said I don't know that it's necessary to do that.

ryan-wallace-frbb: Thanks, that makes total sense. Sorry for over-complicating it.

I went ahead and put the code in the `set` method back in as you suggested.

mcaceresb (approved):

PR #418, opened 2022-04-04
Switch to github actions

kylebarron:
Travis seems pretty dead nowadays.


.github/workflows/test.yml
```diff
@@ -0,0 +1,36 @@
...
+  push:
+    branches:
+      - master
+    tags:
+      - "*"
+  pull_request:
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
```
mcaceresb, line 14: Should this be `runs-on: ${{ matrix.os }}` and under matrix `os: [ ubuntu-latest, windows-latest, macos-latest ]`?

kylebarron: I kept it single-os for now to make this a direct port of the previous travis.yml file. Maybe merge this and then have a new pr to add multiple OS-es?

Issue #416, opened 2022-04-04
Update kernel's python dependencies


mcaceresb:
#### Problem description

The python dependencies, specially those outlined in `requirements_dev.txt`, are dated. Unit tests (via `pytest`) are additionally failing. The issue seems to be that python packages have changed slightly over the years (rather than any fundamental problem with the kernel itself). I tried to update the dependencies and discovered it wasn't particularly simple, hence this issue.

#### Other information

See, for example, #404, #413 (and possibly #409).

NB: Given this wasn't a simple task I may not get to it for some time unless it proves to be a serious issue.

#### TODO

- [ ] Check tests in other OSs.
- [ ] Check rest of tests.

kylebarron:
I now prefer using [Poetry](https://python-poetry.org/) for python projects because it supplies an entire lockfile. I can put up a quick PR if you want to see how it works

PR #419, opened 2022-04-04
Poetry setup

kylebarron:
Ref #416 

A more modern setup with lockfile than using `requirements.txt`. See https://python-poetry.org/

Dependency locking of pywin32 is taking forever though (~30min so far)


mcaceresb (dismissed):

kylebarron:
This branch needs to be updated with master after #418 .

mcaceresb:
@kylebarron This looks good to me; how to modify the workflows to work with poetry?

kylebarron:
I'd suggest using this: https://github.com/snok/install-poetry

So the test ci would become
```yml
      - uses: actions/checkout@v2

      - name: Set up Python ${{ matrix.python-version }}
        uses: actions/setup-python@v2
        with:
          python-version: ${{ matrix.python-version }}

      - name: Set up Poetry
        uses: snok/install-poetry@v1

      - name: Install dependencies
        run: |
          poetry install
          poetry run python -m stata_kernel.install

      - name: Run tests
        # TODO: We should be running other tests and not just these two files
        run: |
          poetry run pytest tests/test_mata_lexer.py tests/test_stata_lexer.py
          # poetry run python setup.py install
```
Not sure off the top of my head how to edit the `setup.py install` step now that we don't have a `setup.py` file.

pyproject.toml
```diff
@@ -0,0 +1,52 @@
+[tool.poetry]
+name = "stata_kernel"
+version = "1.12.3"
+description = "A Jupyter kernel for Stata. Works with Windows, macOS, and Linux."
+authors = ["Kyle Barron <kylebarron2@gmail.com>", "Mauricio Caceres"]
```
kylebarron, line 5: You can edit your email here if you'd like @mcaceresb

pyproject.toml
```diff
@@ -0,0 +1,52 @@
...
+pandas = "^1.0"
+pygments = "^2.2"
+pexpect = "^4.6.0"
+requests = "^2.19.1"
+jupyter-client = "^5.2.3"
+IPython = "^6.5.0"
+ipykernel = "^4.8.2"
+packaging = "^17.1"
+pillow = ">=5.2.0"
+pywin32 = { version = ">=223", platform = "Windows" }
```
kylebarron, line 39: In a separate PR we could update these dependencies. The `^` allows semver compatibility of dependent versions. https://python-poetry.org/docs/dependency-specification/#caret-requirements

kylebarron: So for example the `IPython` dependency above might be problematic because `pip install stata-kernel` would downgrade the user's ipython dependency to `>=6.5,<7`

mcaceresb:
> Not sure off the top of my head how to edit the `setup.py install` step now that we don't have a `setup.py` file. What does `setup.py install` do? I thought it was equivalent to `pip install .`

@kylebarron I think so. My read is that it's not needed for the tests, but I could be wrong.

pyproject.toml
mcaceresb: @kylebarron That's interesting. I agree a separate PR ought to happen after migrating to poetry and gh workflows. I also have to figure out where to test out the dependencies. I think this can be merged once the workflows file is updated.

kylebarron:
> My read is that it's not needed for the tests, but I could be wrong.

I don't think it's needed for the tests. I think it was just to test that the package could be installed

mcaceresb:
@kylebarron 

> > My read is that it's not needed for the tests, but I could be wrong.
> 
> I don't think it's needed for the tests. I think it was just to test that the package could be installed

Is there a command to install a local package with poetry? If yes then that could go in place; if no then the line can just be deleted, right?

kylebarron:
> Is there a command to install a local package with poetry?

This is just `poetry install` from the package root.

